PR #69, opened 2020-05-22
fix(overlay): zindex on backdrop and content

knownasilya:
Not sure why, maybe because the elements are fixed, but the zindex on the parent doesn't promote the children above other elements before and with a lower zindex than the overlay. Saw the issue with ember-leaflet, the map would be over the backdrop and content.


knownasilya:
Any chance for a new release, since I cant use it directly from github due to monorepo?

josemarluedke:
Published in v0.2

PR #70, opened 2020-05-22
Upgrade all project dependencies

josemarluedke:
It seems that we have a few broken tests in ChangesetForm with the latest changes from `ember-changeset`. I will have to investigate it over the next few days.

FYI @snewcomer, once I have cycles, I will try to isolate what is going on. To be fair, I don't know if the issue is from ember-changeset, maybe from the Ember upgrade.


snewcomer:
Another version to try would be 3.4.0.  I'd be curious of a specific test that failed (w/ 3.4) if you could point me to it 👍

josemarluedke:
@snewcomer sorry for the mentioning here. For reference, I pushed a commit with the fixes for Ember Changeset. It seems that there were a few changes that broke a few things:

1. `error.validation` is now a string, and before it was an array of strings.
2. defining a validation for a nested key worked before with `{'something.else': ValidationFunc}` and now it only works when defining as an object: `{ something: { else: validationFunc } }`.

snewcomer:
That looks like the right changes!  Glad all your tests are passing!

I should really highlight these changes in the README so you don't have to go fishing

PR #123, opened 2020-12-02
Notifications: Binding for template usage

betocantu93:
Hello @josemarluedke awesome addon, I was eager to try it out and I'm loving it, its great!

This PR justs adds binding to the notifications service.

I have a preference to use lots queue's and pipe's inside the templates, so for me a normal use case would be to trigger a notification like this

```hbs
<Button {{on "click" (pipe this.oneFn this.secondFn (fn this.notifications.add "all good!")) }}>
  Pipe actions
</Button>
```

Let me know what you think , thanks.

### DIFF
--- a/packages/notifications/addon/services/notifications.ts
+++ b/packages/notifications/addon/services/notifications.ts
@@ -2,6 +2,7 @@ import Service from '@ember/service';
 import Notification from '../-private/notification';
 import NotificationsManager from '../-private/manager';
 import { NotificationOptions } from '../-private/types';
+import { action } from '@ember/object';
 
 export default class NotificationsService extends Service {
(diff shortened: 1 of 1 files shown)


josemarluedke (commented):
Thanks for working on this! 

Changes look good! It seems we have some tests failing due to this change, could you take a look on that?

betocantu93:
Yup, to be honest, I tried to fix the tests, but it seems `@action` decorator have issues with sinon stubs, still, gonna keep digging. If you could point me in the right direction, or have any tip, would be much appreciated

Tried [this approach](https://github.com/elwayman02/ember-sinon-qunit#using-sinon-with-the-action-decorator) without success so far

PR #149, opened 2021-02-19
doc: Overlays + Power Select

### LINKED ISSUE #148: Drawer + Power Select
I have a code that goes something like this (only showing important fields):

```hbs
<Drawer @isOpen={{true}} as |d|>
    <PowerSelect @searchEnabled={{true}} as |el|>
      {{yield el}}
    </PowerSelect>
</Drawer>
```

The problem then is that the `input` search field in `power-select` is not reachable. It renders. I could click it, but it won't get the focus.

My theory is that since the `power-select` itself is rendered through a wormhole and since `drawer` is trying to keep the focus, it just does not work.

I figured out that in my case when I set `@disableFocusTrap={{true}}` on the `drawer` it starts working again. 

Just wanted to bring this issue to attention. Maybe there is a better way? And also to "document" it for future travellers.

Feel free to close if this is "works as intended".

MichalBryxi:
Because power select inserts popover content outside of the drawer
content then any focusable elements will be inaccessible thanks to
focus-trap. This documentation tries to shed some light on how this
can be solved.

Fixes #148

### DIFF
--- a/packages/overlays/addon/components/overlay.md
+++ b/packages/overlays/addon/components/overlay.md
@@ -17,7 +17,7 @@ It contains all the core features necessary for a great experience.
 `FormSelect` uses Power Select under the hood and has `@renderInPlace={{false}}` set by default, which will insert the content of the drop-down outside of the `Overlay` (or `Drawer` or `Modal`) and because of that focus-trap will prevent accessing focusable elements ([search input](https://ember-power-select.com/docs/the-search)) in said Power Select. To solve this you have following options:
 
 - `<Overlay @disableFocusTrap={{true}} />` (not recommended)
-- `<FormSelect @renderInPlace={{false}} />` (recommended)
+- `<FormSelect @renderInPlace={{true}} />` (recommended)
 
 ## API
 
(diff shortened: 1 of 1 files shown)


MichalBryxi:
Feel free to correct my potato-english.

packages/overlays/addon/components/overlay.md
```diff
@@ -12,6 +12,13 @@ It contains all the core features necessary for a great experience.
 
 > Important: You must have a focusable element inside of the Overlay.
 
+## Note on Overlays and other portal elements
+
+`FormSelect` uses Power Select under the hood and has `@renderInPlace={{false}}` set by default, which will insert the content of the drop-down outside of the `Overlay` (or `Drawer` or `Modal`) and because of that focus-trap will prevent accessing focusable elements ([search input](https://ember-power-select.com/docs/the-search)) in said Power Select. To solve this you have following options:
+
+- `<Overlay @disableFocusTrap={{true}} />` (not recommended)
+- `<FormSelect @renderInPlace={{false}} />` (recommended)
```
josemarluedke, line 20: Should this be `true`? 

```suggestion
- `<FormSelect @renderInPlace={{true}} />` (recommended)
```

MichalBryxi: true...

josemarluedke:
Thank you @MichalBryxi!

PR #152, opened 2021-04-14
fix(overlays): set modal/drawer body to overflow auto

knownasilya:
Prevents an ugly scroll bar that's always there.

Screen Shot 2021-04-14 at 3 24 17 PM


josemarluedke:
@knownasilya Excellent! Can you make the same change for the Drawer component?

knownasilya:
@josemarluedke all set

josemarluedke:
@knownasilya Thank you.

PR #155, opened 2021-04-16
Test Embroider compatibility in CI

josemarluedke:
Also moves CI to use volta action.


josemarluedke:
It seems that the Embroider compat layer for ember-power-select is out of date, which causes the embroider-optimized case to fail:

```
Build Error (PackagerRunner) in ../../../../node_modules/ember-power-select/components/power-select/options.hbs

Module Error (from /Users/josemarluedke/code/oss/frontile/node_modules/thread-loader/dist/cjs.js):
Unsafe dynamic component: @groupComponent in $TMPDIR/embroider/7da432/node_modules/ember-power-select/components/power-select/options.hbs/options.hbs


```

PR #173, opened 2021-10-28
Add optional validation on init and allow forcing field error visibility through the form

bitwolfe:
We needed the ability to show validation errors immediately when rendering the form. For this I added a property to ChangesetForm called alwaysShowErrors, which when true sends this flag on to the fields by piggybacking on their existing hasError property, which apparently only did anything if it was false.

This way you can tell the form to always render field errors when needed and not just after field focus out or submit.

Our specific case for this is when running with our own choice of validation library with validations that should vary depending on other fields. So if a required field becomes visible after checking a checkbox the user gets told that the new field is also required, without them first having to try and submit the form.

Thought I would make a PR in case you think this would be a good addition to the addon. Feedback is welcome!


bitwolfe:
@josemarluedke Please have a look again when you get the chance. I also figured out the docs and regenerated the args data so the new options should show up.

nanderss:
@josemarluedke Would it be possible to get this merged?

nanderss:
@josemarluedke Could we get this merged please?

josemarluedke (approved):
Changes look good! 🎉

josemarluedke:
@bitwolfe Thanks for working on this! 🎉 🎉

PR #194, opened 2022-02-02
fix(#176): tailwind v3.x safelisting

### LINKED ISSUE #176: purge safelist won't work with JIT
The [safelist](https://frontile.dev/docs/installation#using-the-safelist-config-option) won't really [work with JIT](https://v2.tailwindcss.com/docs/just-in-time-mode#limitations), because:

> The safelist option does not support regular expressions. Because no CSS is generated by default, the safelist has to be a list of complete class names. It’s not possible to safelist a regular expression, because there is not a pre-generated list of class names to match against that regular expression.

Or from [v3.x documentation](https://tailwindcss.com/docs/content-configuration#safelisting-classes).

MichalBryxi:
The [safelist](https://frontile.dev/docs/installation#using-the-safelist-config-option) won't [work with JIT](https://v2.tailwindcss.com/docs/just-in-time-mode#limitations) the way it worked before, because:

> The safelist option does not support regular expressions. Because no CSS is generated by default, the safelist has to be a list of complete class names. It’s not possible to safelist a regular expression, because there is not a pre-generated list of class names to match against that regular expression.

From tailwind [v3.x the documentation hints](https://tailwindcss.com/docs/content-configuration#using-regular-expressions) that `safelist` using regular expressions matching would work.

Fixes #176

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -121,7 +121,7 @@ There are two practical approaches to set up PurgeCSS to prevent the erasing of
 all styles provided by Frontile. The simplest way is to use magical comments to
 safelist the TailwindCSS components. The other way is to list all the classes
 (or at least their prefix) in the PurgeCSS safelist config.
-Check [PurgeCSS docs](https://purgecss.com/safelisting.html) for further info on safelisting.
+Check [PurgeCSS docs](https://purgecss.com/safelisting.html) adn [tailwindcss docs](https://tailwindcss.com/docs/content-configuration#using-regular-expressions) for further info on safelisting.
 
 ### Using comments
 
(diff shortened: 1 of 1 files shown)


MichalBryxi:
I'm not sure whether it's responsibility of this addon to educate users to put respective paths into `content` key of tailwind config, as all their styles would be purged without it anyway. But to give a hint:

```js
'use strict';
// located in <app root>/config/tailwind/

const path = require('path');

const appRoot = path.join(__dirname, '../../');
const appEntry = path.join(appRoot, 'app');
const relevantFilesGlob = '**/*.{html,js,ts,hbs,gjs,gts}';

module.exports = {
  content: [path.join(appEntry, relevantFilesGlob)],
  safelist: [ ... ]
}
```

josemarluedke (approved):

josemarluedke:
@MichalBryxi Thanks for the PR. 

I think that could be a section in the documentation where we give another option instead of safelist. If we set content and include `node_modules/@frontile`, that would work as well.

PR #204, opened 2022-03-22
Upgraded packages to use tailwindcss 3

eugenioenko:
- Upgraded talwindcss to v3
- Updated visuallyHidden component to use sr-only class
- removed purgecss
- updated talwind.config.js to include [content]

---- 

## Breaking Change

The core pkg no longer ships the `visually-hidden` and `js-focus-visible` styles. These were previous provided by an css file that Ember CLI would import into your vendor file. These styles are now provided by `@frontile/core` tailwind plugin. 

You also will need to add to your tailwind safelist the following new rules:

```js
{ pattern: /^js-focus-visible/ },
{ pattern: /^sr-only/ },
```

You can remove the following:

```js
{ pattern: /^visually-hidden/ },
```

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -128,9 +128,7 @@ File `app/styles/app.css`:
 
 ```css
 @tailwind base;
-/* purgecss start ignore */
 @tailwind components;
-/* purgecss end ignore */
 @tailwind utilities;
 ```
 
@@ -142,6 +140,7 @@ File `app/styles/app.css`:
 module.exports = {
   safelist: [
     { pattern: /^close-button/ },
+    { pattern: /^js-focus-visible/ },
     { pattern: /^sr-only/ },
 
     // Frontile Forms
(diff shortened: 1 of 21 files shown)


eugenioenko:
I'm still getting this warning  even thought technically `content:` is set
image

site/app/config/tailwind.config.js
```diff
@@ -144,6 +144,7 @@ module.exports = {
   ],
   safelist: [
     { pattern: /^close-button/ },
+    { pattern: /^js-focus-visible/ },
```
josemarluedke, line 147: Can you also add this to the docs?

docs/installation.md
```diff
@@ -143,7 +140,8 @@ File `app/styles/app.css`:
 module.exports = {
   safelist: [
     { pattern: /^close-button/ },
-    { pattern: /^visually-hidden/ },
+    { pattern: /^js-focus-visible/ },
```
eugenioenko, line 143: Docs updated

PR #205, opened 2022-03-24
Enable embroider safe for tests; Upgrade dependencies

eugenioenko:


packages/notifications/package.json
```diff
@@ -25,29 +25,30 @@
     "@ember/render-modifiers": "^2.0.4",
     "@frontile/core": "^0.13.1",
     "@frontile/tailwindcss-plugin-helpers": "^0.13.0",
+    "ember-auto-import": "^2.4.1",
```
eugenioenko, line 28: so `ember s` requested that `ember-auto-import` be moved as a dependency  and not devDependency so that the `ember -css-transitions` work.

josemarluedke: Sounds good.

PR #206, opened 2022-03-24
Enable embroider optimized and Drop ChangesetForm::Context

eugenioenko:
Breaking changes:
- `ChangesetForm` component no longer yields the changeset obj as a second param. 
- `ChangesetForm::Context` component has been removed.


packages/changeset-form/addon/components/changeset-form/index.hbs
```diff
@@ -3,21 +3,54 @@
   {{on "submit" (fn this.handleSubmit @changeset)}}
   {{on "reset" (fn this.handleReset @changeset)}}
 >
-  <ChangesetForm::Context @changeset={{@changeset}} as |Form|>
```
eugenioenko, line 6: Two different fixes here:

**Unsafe dynamic component**:
embroider optimized doesn't like the re-yielding of dynamic components, its throws the error `Unsafe dynamic component` when using `ChangesetForm::Context`.
So replacing `Form.Input` with the component path solves the issue.

**Yielding a list**
So it looks like doing a `{{yield one two three}}` is okey, but when the component is used as `<MyComponent as |one two three| /> the second and third binding is undefined.

Switched to using a single hash yield. So the `changeset` now would be `Form.changeset` on the component user. Thats a breaking change

PR #213, opened 2022-05-25
Cursor pointer (hand) on radios/checkbox + label

eugenioenko:
- Added enabled checkbox/radio inputs cursor  `pointer` (hand).
- Added disabled checkbox/radio input cursor `not-allowed`
- Added enabled checkbox/radio label cursor `pointer`
- Added disabled checkbox/radio label cursor `not-allowed`

Note: 
Before this pr the structure was:
```
<label-container>
    <input-container>
       <input />
   </input-container>
   <label/>
<label-container>
```
Because the child `:has()` pseudo selector is not yet available, and the html `...attributes`  go for the `input` element, the only way to make cursor apply to the label only when input is not disabled  is to have the `label` be a sibling of `input` and use the  adjacent sibling combinator selector `+`. Something like `input:not([disabled]) + label`

 For that, `label-container` container needed to be removed.

So the new structure is 
```
<input-container>
   <input />
   <label/>
</input-container>
```

At simple glance, the extra `label-container` wrapper doesn't seems to be exactly necessary, but of course, I might be missing something. Is it safe to remove? 
image

### DIFF
--- a/packages/forms/tailwind/default-options.js
+++ b/packages/forms/tailwind/default-options.js
@@ -221,14 +221,22 @@ function defaultOptions({ config }) {
       },
 
       parts: {
-        labelContainer: {
+        inputContainer: {
           lineHeight: defaultTheme.lineHeight.tight,
           display: 'flex',
-          alignItems: 'flex-start'
-        },
-        inputContainer: {
-          display: 'flex',
-          alignItems: 'center'
+          alignItems: 'center',
+          '& > input:not([disabled])': {
+            cursor: defaultTheme.cursor['pointer'],
+            '& + label': {
+              cursor: defaultTheme.cursor['pointer']
+            }
+          },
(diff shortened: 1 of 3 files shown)


packages/forms/tailwind/default-options.js
```diff
@@ -221,14 +221,22 @@ function defaultOptions({ config }) {
       },
 
       parts: {
-        labelContainer: {
-          lineHeight: defaultTheme.lineHeight.tight,
-          display: 'flex',
-          alignItems: 'flex-start'
-        },
         inputContainer: {
+          lineHeight: defaultTheme.lineHeight.tight,
```
eugenioenko, line 225: `lineHeight: defaultTheme.lineHeight.tight` 

not sure if this should be on the label or  here

PR #229, opened 2023-05-07
Mark radio and checkbox container elements as checked to allow label styling

jkeen:
Currently with the `<input>` element inside a wrapper element, you can't style the form labels differently based on the checked state of the input (which I'd normally do with a `input:checked + label` selector)

This PR adds a `--checked` variant class to the top-level form elements of both the radio buttons and checkboxes when they're checked so that the child elements can be targeted and styled

### DIFF
--- a/packages/forms/addon/components/form-checkbox.hbs
+++ b/packages/forms/addon/components/form-checkbox.hbs
@@ -11,7 +11,12 @@
     class={{use-frontile-class "form-checkbox" @size part="label-container"}}
   >
     <div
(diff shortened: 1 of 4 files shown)


jkeen:
@josemarluedke Josemar! I've slowly been migrating my some old custom components to using frontile, and even without being a tailwind guy this component set is saving me so much time and headache. Thanks for all your work on this beautiful library

josemarluedke:
The changes look good. Thanks for working on them.

PR #230, opened 2023-05-07
Add ids and events to notifications

jkeen:
I wanted to be able to mark a notification record as read on the backend when a user closed a (frontile) notification on the frontend, and this PR opens up that possibility.

- Adds an `id` attribute to the notification object in frontile, and that `id` is output as a `data-notification-id` attribute on the Notification component. 
- When a notification is added or removed on the frontend, an event is fired on the manager object, allowing consumers to tie into that. 

An example of this in use: 

```javascript
...
  @service cable;
  @service notifications

  subscribeToUpdates(user) {
    this.subscribeToNotificationsChannel(user);

    this.notifications.manager.on('remove', this.notificationRead.bind(this));
  }

  notificationRead(notification) {
    if (this.subscriptions.notificationsChannel) {
      this.subscriptions.notificationsChannel.perform('mark_as_read', {
        ids: [notification.id],
      });
    }
  }
...

```

### DIFF
--- a/packages/notifications/addon/-private/evented.ts
+++ b/packages/notifications/addon/-private/evented.ts
@@ -1,0 +1,38 @@
+import EmberObject from '@ember/object';
+import EmberEvented from '@ember/object/evented';
+
+/**
+ * Partial wrapper for Ember's Evented Mixin, enabling
+ * a pure class-based derivation
+ * @private
+ */
+
+export default class Evented {
+  eventManager = EmberObject.extend(EmberEvented).create();
+
+  on(name: string, target: object, method?: string | Function): this {
+    //@ts-ignore
+    this.eventManager.on(name, target, method);
+    return this;
+  }
+
+  off(name: string, target: object, method?: string | Function): this {
+    //@ts-ignore
+    this.eventManager.off(name, target, method);
+    return this;
+  }
+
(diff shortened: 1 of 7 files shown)


josemarluedke:
I've taken a different approach to implementing this functionality that avoids the Evented pattern while still supporting the core use cases you outlined.

Instead of using events, I implemented a direct callback system that's simpler and more predictable.

Key differences in the new approach:
- Single callback: onDismiss callback on NotificationsContainer instead of separate 'add'/'remove' events
- Generic metadata support: Type-safe metadata with TypeScript generics
- No Evented pattern: Direct callback invocation for better performance and simpler debugging
- Unified dismissal handling: Callback fires for both manual dismissal and timer expiry

Your use case example:
// Your original approach
this.notifications.manager.on('remove', this.notificationRead.bind(this));

// New approach
<NotificationsContainer @onDismiss={{this.notificationRead}} />

notificationRead = (notification) => {
  if (notification.metadata?.notificationId) {
    this.api.markNotificationAsRead(notification.metadata.notificationId);
  }
}

The new implementation is available in PR #375: https://github.com/josemarluedke/frontile/pull/375

This provides all the functionality you requested (backend integration, unique IDs via metadata, event tracking) but with a cleaner API that's easier to use and maintain. Would love to get your feedback on this approach!

PR #298, opened 2024-06-26
Adds push dist so a developer can point to the main branch of this repo

cloke:
With the v2 add-on format we can't easily point to the main branch of an add-on. The builds a dist branch after merges similar to other v2 add-ons.


josemarluedke:
Does that work well with the monorepo setup?

Issue #323, opened 2024-11-24
Live demo in ChangesetForm documentation page is broken


MichalBryxi:
- See: https://frontile.dev/docs/changeset-form/usage/
- Page seems to render ok, but the live demo does not render

The error reads:

```
Error: Could not find module `@ember/string` imported from `ember-changeset-validations/utils/messages`
```

PR #325, opened 2024-11-24
fix: Unexpected identifier 'assert'

MichalBryxi:
- Without this hotfix the `pnpm build` of a `packages/theme` would fail with:
```
[js] [!] SyntaxError: Unexpected identifier 'assert'
[js]     at compileSourceTextModule (node:internal/modules/esm/utils:340:16)
[js]     at ModuleLoader.moduleStrategy (node:internal/modules/esm/translators:102:18)
[js]     at ModuleLoader.#translate (node:internal/modules/esm/loader:433:12)
[js]     at ModuleLoader.loadAndTranslate (node:internal/modules/esm/loader:480:27)
```
- This is likely because: `browserslist-generator@v3.0.0` [adds compatibility](https://github.com/wessberg/browserslist-generator/blob/master/CHANGELOG.md) for:
```
bump dependencies and support Node 22
```
- Fixes #324

PR #326, opened 2024-11-24
fix: Could not find module `@ember/string` imported from `ember-changeset-validations/utils/messages`

MichalBryxi:
- Without this fix [changeset-form](https://frontile.dev/docs/changeset-form/usage) is broken because of:
```
Error: Could not find module `@ember/string` imported from `ember-changeset-validations/utils/messages`
```
- Fixes [#323](https://github.com/MichalBryxi/frontile/issues/323)

#Screenshot

Now the live example renders:

Screenshot 2024-11-24 at 18 44 45


MichalBryxi:
This PR contains also the commits from: https://github.com/josemarluedke/frontile/pull/325

If we merge that one first, this one will get smaller. IDK if there is a simpler git flow.

josemarluedke (approved):

PR #388, opened 2025-08-29
feat: Table component

josemarluedke:


josemarluedke:
Thanks for pointing that out, @MichalBryxi. I like your suggestion and pushed a few commits to start moving toward that dependency. There’s still more to do, but this is a good first step.

PR #395, opened 2025-09-26
chore: add `isLoading` to form on submit

rmoreyunderline:
Add an `isLoading` boolean yielded by the form component.  The `isLoading` boolean is `false` by default.  It is set to `true` during execution of the `onSubmit` argument, which may be a promise.  After `onSubmit` resolves, rejects, or completes synchronously, `isLoading` is set back to `false`.  New test case included.

Example:
 ```hbs
 <Form
   @onSubmit={{this.onSubmit}}
   as |form|
 >
   <input name="firstName" />
   <input name="lastName" />
   <button type="submit" disabled={{form.isLoading}}>
     {{#if form.isLoading}}Submitting...{{else}}Submit{{/if}}
   </button>
 </Form>
 ```

### DIFF
--- a/packages/forms/src/components/form.gts
+++ b/packages/forms/src/components/form.gts
@@ -20,14 +20,14 @@ interface FormSignature {
     /**
      * Optional callback invoked on input changes within the form.
      */
-    onChange?: (data: FormResultData, event: Event) => void;
+    onChange?: (data: FormResultData, event: Event) => Promise<void> | void;
     /**
      * Callback invoked on form submission.
      */
     onSubmit: (
       data: FormResultData,
       event: SubmitEvent
-    ) => void | Promise<void>;
+    ) => Promise<void> | void;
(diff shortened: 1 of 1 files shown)


packages/forms/src/components/form.gts
```diff
@@ -1,26 +1,74 @@
...
+  isLoading: boolean;
+}
+
 interface FormSignature {
   Element: HTMLFormElement;
   Args: {
+    /**
+     * Optional callback invoked on input changes within the form.
+     */
     onChange?: (data: FormResultData, event: Event) => void;
```
josemarluedke, line 23: We probably should change the return type to be `Promise<void> | void`.

When we get to validation, we could also allow it to return a list of errors from backend validations, but we can discuss that when we reach that part.

josemarluedke (approved):

PR #396, opened 2025-09-29
feature: standard schema validation utility

rmoreyunderline:
Add a `StandardValidator` utility to the forms package which provides validation conveniences for schemas in [Standard Schema](https://standardschema.dev/) form.

`StandardValidator` provides the following conveniences:
* Validates arbitrary data against a standard schema.
* Validates single fields against a standard schema.
* Validates arbitrary data against a custom validator function.
* Validates single fields against a custom validator function.
* Validates against either/or schema and custom validator, if desired.
* Typed, with a basis in Standard Schema.

### DIFF
--- a/packages/forms/package.json
+++ b/packages/forms/package.json
@@ -75,6 +75,7 @@
     "@glint/environment-ember-template-imports": "1.5.2",
     "@rollup/plugin-babel": "^6.0.4",
     "@rollup/plugin-node-resolve": "^16.0.0",
+    "@standard-schema/spec": "^1.0.0",
     "@tsconfig/ember": "^3.0.9",
     "babel-plugin-ember-template-compilation": "^2.2.1",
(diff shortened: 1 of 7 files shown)


josemarluedke (commented):
Changes are looking good. I would like to see some additional tests to ensure that we can work effectively with objects inside other objects.

```ts
const registrationSchema = v.object({
  something: v.pipe(v.string()),
  personalInfo: v.object({
    firstName: v.pipe(v.string(), v.nonEmpty('First name is required')),
    lastName: v.pipe(v.string(), v.nonEmpty('Last name is required')),
    birthDate: v.pipe(v.string(), v.isoDate('Invalid date format')),
  }),
});
```

Then, when we call `filterFieldIssues('personalInfo.firstName')` it works...

josemarluedke (approved):

PR #398, opened 2025-10-01
feature:  automatic form validation

rmoreyunderline:
Form now accepts a standard schema and validates on submit.  The new `Field` component acts as an intermediary between `Form` and form elements to auto-bind any error messages from the validation.


josemarluedke (approved):
🎉 This looks very good. I left a few inline comments, but these are not blocking.

PR #404, opened 2025-10-13
feat: Form `validateOn` for optional `change`, `input`, and `submit` validations

rmoreyunderline:
This PR introduces flexible validation timing controls to the Frontile forms library through the `validateOn` argument, enabling developers to specify when form validation should occur across three key moments: field changes (blur events), user input (keystroke events), and form submission. The Form component now accepts `validateOn` with options `'change'`, `'input'`, and/or `'submit'` (defaulting to `['change', 'submit']`), which automatically cascades to all Field components, while individual Fields can override this setting for fine-grained control over validation UX—such as enabling real-time input validation for password strength indicators while keeping other fields on blur-based validation. The implementation includes debounced input validation to optimize performance, comprehensive test coverage for all validation timing scenarios including dynamic changes and per-field overrides, and maintains backward compatibility by preserving the default behavior of validating on both field changes and form submission.

**Note**:  blur validation will be implemented in a separate PR.

```ts
import { Form } from '@frontile/forms';
import * as v from 'valibot';

const schema = v.object({
  username: v.pipe(v.string(), v.minLength(3, 'At least 3 characters')),
  password: v.pipe(v.string(), v.minLength(8, 'At least 8 characters'))
});

<template>
  <Form 
    @schema={{schema}} 
    @validateOn={{array 'change' 'submit'}} 
    @onSubmit={{this.handleSubmit}} 
    as |form|
  >
    {{! Username validates on change/blur }}
    <form.Field @name="username" as |field|>
      <field.Input @label="Username" />
    </form.Field>

    {{! Password validates on every keystroke (overrides Form setting) }}
    <form.Field @name="password" @validateOn={{array 'input'}} as |field|>
      <field.Input @label="Password" @type="password" />
    </form.Field>

    <button type="submit">Register</button>
  </Form>
</template>
```

### DIFF
--- a/packages/forms/src/components/field.gts
+++ b/packages/forms/src/components/field.gts
@@ -101,16 +101,6 @@ class Field<
     return this.args.validateOn ?? ['change'];
   }
 
-  /** Whether validation should run on field change. */
-  get validateOnChange(): boolean {
-    return this.validateOn.includes('change');
-  }
-
-  /** Whether validation should run on field input. */
-  get validateOnInput(): boolean {
-    return this.validateOn.includes('input');
-  }
-
   /**
    * Returns the current value for the field from formData.
    * Supports both flat and dotted field names (e.g., 'email' or 'profile.email').
@@ -123,7 +113,9 @@ class Field<
     return get(this.args.formData, this.args.name);
   }
 
-  @action
+  /**
+   * Validates the field by calling the validateField function passed in args.
+   */
   validateField() {
     this.args.validateField?.(this.args.formData as T, this.args.name);
   }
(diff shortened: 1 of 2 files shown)


packages/forms/src/components/field.gts
```diff
@@ -103,12 +123,31 @@ class Field<
     return get(this.args.formData, this.args.name);
   }
 
+  @action
```
josemarluedke: `@action` decorator is not needed here as this method is not called from a template.

rmoreyunderline:
```suggestion
```

PR #407, opened 2025-10-24
feat: Select component bidirection state management

rmoreyunderline:
### Feature: Select Component - Bidirectional State Management

This PR fixes the Select component to properly support both internal and external state management, enabling reliable two-way data binding patterns.

#### Problem
The Select component's state management was inconsistent making it difficult to maintain or reason about.  Previously, state management depended on whether or not `@onSelectionChange` was set.

#### Solution

**The new data flow system provides a cleaner, more predictable state management model:**

Previously, the component had conditional logic that decided whether to display values from the parent component or manage its own state, based on whether a callback was provided. This created an "either-or" situation where you couldn't have both parent-controlled state AND a responsive UI.

The new approach **eliminates this conditional behavior entirely**. Now the component remains synchronized with parent-provided values regardless if a change handler is passed. This creates a much simpler mental model:

1. **Always Synchronized**: When parent values change, the component's display updates automatically
2. **Immediate Feedback**: When users make selections, the UI updates instantly while the parent is optionally notified
3. **Predictable Cycle**: Parent updates can optionally flow back into the component.

Now the UI always displays the correct values, whether changes originate from user clicks or from parent component updates.

#### Changes
- **[select.gts](packages/forms/src/components/select.gts)**: Refactored state management logic and added comprehensive JSDoc comments explaining the data flow architecture
- **[field-test.gts](test-app/tests/integration/components/forms/field-test.gts)**: Added Field integration tests (6 new tests)
- **[select-test.gts](test-app/tests/integration/components/forms/select-test.gts)**: Added comprehensive state management tests (18 new tests)

#### Tests Added (24 total)
**Field Integration Tests (6)**:
- SingleSelect with formData: initialization and external updates
- MultiSelect with formData: initialization, external updates, user interactions with/without initial data

**Select State Management Tests (18)**:
- Single mode: external selectedKey changes (trigger text, native select, listbox sync)
- Single mode: user selection updates external state
- Single mode: rapid external changes with user interaction
- Multiple mode: external selectedKeys changes (trigger text, listbox sync)
- Multiple mode: user selections update external state
- Multiple mode: adding/removing selections with external state
- Multiple mode: rapid external changes with user interaction
- Edge cases: clearing selections, array reference changes, empty selections

### DIFF
--- a/packages/forms/src/components/select.gts
+++ b/packages/forms/src/components/select.gts
@@ -371,9 +371,9 @@ interface SelectSignature<T> {
  *    - Parent updates its state, which flows back through step 1
  *
  * ### Why Both Internal State AND Callbacks?
- * - **Internal state:** Enables immediate, reactive UI updates (`_selectedKey` / `_selectedKeys`)
- * - **Parent callback:** Enables parent to control and validate changes (`@onSelectionChange`)
- * - This pattern provides both responsive UX and parent control over state
+ * - **Internal state:** Enables immediate, responsive UI updates (`_selectedKey` / `_selectedKeys`)
+ * - **Parent callback:** Enables parent to take action when changes occur (`@onSelectionChange`)
+ * - This pattern provides both responsive UX and parent inclusion over state
  *
  * ### Example Flow
  * ```


packages/forms/src/components/select.gts
```diff
@@ -374,14 +520,26 @@ class Select<T = unknown> extends Component<SelectSignature<T>> {
...
+   */
   get selectedKeys(): string[] {
     if (this.args.selectionMode === 'multiple') {
-      if (
-        typeof this.args.selectedKeys !== 'undefined' &&
-        typeof this.args.onSelectionChange === 'function'
-      ) {
-        return this.args.selectedKeys;
-      }
+      // if (
```
josemarluedke: We probably can remove this commented code.

PR #500, opened 2026-08-21
feat(site): prerender all docs routes with Vite SSR + Glimmer rehydration

josemarluedke:
Every Docfy route is now rendered to static HTML at build time, and the client **rehydrates** that DOM instead of rendering a second copy over it.

- **68 routes prerender in ~2s**; `pnpm --filter site build` (the command Netlify runs) takes ~32s end to end.
- Client JS is 5.6 MB raw / **905 KB gzipped in a single chunk** — that's the point: readable content now paints before any of it parses.
- Serialize markers add 27–37% to raw HTML but compress away. Worst page (Table) is 621 KB raw → **37 KB gzipped**.

Full write-up, numbers, and known gaps: [`site/ssr/README.md`](https://github.com/josemarluedke/frontile/blob/claude/ssr-vite-research-817957/site/ssr/README.md).

## Why not FastBoot

`ember-cli-fastboot` is a v1 addon needing its own broccoli build tree; that was never ported to Embroider + Vite ([embroider#112](https://github.com/embroider-build/embroider/issues/112), [embroider#2035](https://github.com/embroider-build/embroider/issues/2035)). Ember core's stated path is to deprecate it in favour of Vite's own SSR, which is what this does.

Nothing in `ember-source` had to be worked around. Ember 6.6 still supports server rendering via `ApplicationInstance` boot options — `isBrowser: false`, a custom `document`, `rootElement`, and the private `_renderMode`. FastBoot was only ever a wrapper around those.

## Why it's cheap for this app

Every addon the site depends on is a **v2 addon** (only `ember-auto-import` is v1, and it's build-time only). So `dist/@embroider/virtual/vendor.js` is **316 bytes** of `EmberENV` config with no AMD payload — no `vm` sandbox, no script evaluation ordering. `vite build --ssr` bundles the app for Node as-is.

## Rehydration is the load-bearing part

Ember **appends** to `rootElement` rather than clearing it, so prerendered HTML plus a normal boot renders the whole app twice (verified: two `<h1>`s, duplicated navs).

- Server renders with `_renderMode: 'serialize'` → `<!--%+b:0%-->` boundary markers.
- Client boots with `_renderMode: 'rehydrate'` → adopts the existing DOM.

`_renderMode` is unreachable under autoboot (`didBecomeReady` calls `instance._bootSync()` with no arguments), so `index.html` switches to `autoboot: false` + `app.visit()` when it sees the prerender marker meta tag. **The dev-server path is untouched** and was verified unchanged.

## Bug fixes needed to get here

- **`packages/frontile/src/utils/nested-data.ts`** did a bare `value instanceof FileList`. `FileList` doesn't exist in Node, so this was a hard `ReferenceError` rather than `false` — it broke *any* attempt to server-render a Frontile `Form`, not just this site. Now guarded with `typeof`.
- **`site/app/components/version-dropdown.gts`** and **`docfy-copy-page.gts`** read `window.location.origin` during render. Both now go through a new `currentOrigin()` helper that falls back to `config.siteURL` (overridable via `SITE_URL`), so prerendered HTML is *correct*, not merely crash-free. The real origin takes over on rehydration — which is what keeps the version dropdown working on the `next` subdomain.

Both source fixes are load-bearing: the prerenderer needs **no** `FileList` or `location` shims.

## Build wiring

`site/package.json` now chains the pipeline, so the existing Netlify command needs no change:

```
build        → build:client && build:ssr && prerender
build:client → vite build                                # dist/
build:ssr    → vite build --config vite.config.ssr.mjs   # dist-ssr/
prerender    → node ssr/prerender.mjs                    # 68 × index.html into dist/
```

A failed route sets a non-zero exit code, so a prerender regression fails the build instead of silently falling through to the SPA shell.

`netlify.toml` needs no change either: per Netlify's docs, "By default, you can't shadow a URL that actually exists within the site", so a non-forced `/* -> /index.html 200` rewrite only fires for paths with no matching file. Prerendered pages win; the SPA fallback still covers the `not-found` catch-all.

## Verification

- Full test suite: **829 pass, 3 skip, 0 fail**.
- `pnpm --filter frontile lint:types` clean; all changed files lint and Prettier clean. (`site lint:types` has 2837 pre-existing errors in Docfy-generated demo templates — unrelated, and none in the files touched here.)
- Browser sweep over `/`, table, select, modal, form, autocomplete, notifications, avatar: one `h1` each, zero leftover rehydration markers, **zero console errors**, routing and interaction live.
- `Form` specifically: typed into a field post-rehydration, submitted, and confirmed the value flows through the `nested-data` path into the rendered output.
- Dev server (`pnpm start`) verified unaffected.

## Left as follow-ups

- **`<meta name="description">` is still the generic shell one on every page.** Titles are per-page (ember-page-title runs server-side fine); descriptions aren't. Biggest remaining SEO win, and the Docfy frontmatter is right there.
- No canonical URL or per-page Open Graph tags.
- Worth confirming trailing-slash resolution on the deploy preview, i.e. that `/docs/components/buttons/button` serves `.../button/index.html`.

Draft because the deploy preview is the real test for the two deployment assumptions above.


josemarluedke:
The deploy preview earned its keep — it caught a real regression before merge.

Prerendering was writing `<route>/index.html`, which Netlify only serves at `/route/`. But **every link the site renders is extensionless and slash-free**, including Docfy index pages whose URLs *do* carry a trailing slash:

```
$ curl -sI https://deploy-preview-500--frontile-next.netlify.app/docs/components/buttons/button
HTTP/2 301
location: /docs/components/buttons/button/
```

So every cold page load took a redirect hop, and the site's canonical URLs silently moved. Fixed in 074d322a by emitting flat `<route>.html` files, which Netlify's Pretty URLs serve directly at the existing URLs.

Verified locally that the layout is right — `docs/get-started.html` and `docs/get-started/installation.html` coexist, and root is the only `index.html`. The next preview should show 200s with no `location` header on the non-slash URLs.

Same commit also adds `site/ssr/*.mjs` and `site/vite.config.ssr.mjs` to the root `.eslintrc.js` Node-`.mjs` allowlist, which is what the Lint JS failure was about (`process`/`console` undefined). My earlier local check used the site's flat config; CI runs the root eslintrc, which has the explicit allowlist.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #510, opened 2026-08-26
refactor(forms): decompose Select into a select/ directory

josemarluedke:
## Summary

Behaviour-preserving decomposition of `Select`, which had reached 1387 lines and become a defect generator. Follow-up to #509 (now merged); this diff contains only the restructuring.

`select-test.gts` is **not modified by this PR**. Its 921 lines of behavioural assertions pass unchanged against a fully restructured component, which is the main evidence that behaviour is preserved.

## Why

During #509, four implementers each introduced a defect *elsewhere* in this file while making a localised change (placeholder gating, orphaned JSDoc, a runtime dependency on a test attribute), and reviewers could not hold enough of it in view to notice the trigger had collapsed to `height: 0` and was unclickable. The file's size had become the cause.

## Structure

Follows the existing `collections/listbox.ts` + `listbox/` precedent.

| File | Owns | Lines |
| --- | --- | --- |
| `forms/select.ts` | barrel — public import path unchanged | 2 |
| `select/select.gts` | args wiring, selection sync, orchestration | 831 |
| `select/types.ts` | arg interfaces (464 lines, 137 of code — rest is API-table JSDoc) | 464 |
| `select/trigger.gts` | button vs filterable input | 119 |
| `select/selected-items.gts` | chips / text rendering of the selection | 109 |
| `select/native-mirror.gts` | the hidden native `<select>` — **one copy** | 99 |
| `select/end-content.gts` | spinner / clear / chevron | 78 |

For calibration: `table/table.gts` is 925 lines and orchestrates its own 11-file directory, so `select.gts` at 831 is now in line with the repo's norm rather than an outlier.

## The three concept fixes

1. **One selection projection.** `selectedItems` and `selectedTextValue` were independently walking the node list to answer the same question. During #509 a truncation bug had to be fixed in both and was only caught in the second by a reviewer. `selectedTextValue` now derives from `selectedItems`. A third, vestigial projection (`selectedText`, over keys rather than labels) is deleted — it had no callers.
2. **One `allowEmpty` rule.** `chipsRemovable` reimplemented the deselect rule `ListManager.toggleSelectedItem` already owned. Both now call one function, so the dropdown and a chip's close button cannot disagree about whether the same action is permitted.
3. **The native mirror is one copy.** It was written out twice — once per selection mode — with character-identical `:item` and `:default` blocks. That duplication produced a real bug in #509: the form-value truncation fix had to be applied to both arms, and the single-mode arm had carried the identical defect. The two modes differ in exactly three args, which are now curried onto `(component NativeSelect …)`.

## Organised for the next feature (not implemented here)

Rendering arbitrary content per selected item. `ListItem` now remembers the item it was registered from, so `selectedItems` yields `{ key, textValue, item }` instead of text alone — previously that feature was structurally unbuildable, since the projection came from DOM nodes that never held the source object. Landing it should mean adding a `<:selectedItem>` block in `selected-items.gts` and nothing in `select.gts` or `ListManager`.

A block-form (declarative) option yields `undefined` for `item` — nothing is invented, and that is pinned by a test.

## Invariants deliberately preserved

Each of these was a real bug found by a user or reviewer in #509, and each is still covered:

- the trigger keeps a hittable box (geometry test uses `elementFromPoint`, not `click()`);
- the trigger shares a flex line with the chips rather than wrapping below them;
- `{{p.anchor}}` / `{{p.measureWidth}}` stay on the chips wrapper — on `innerContainer` they abort the module with a `ResizeObserver` loop;
- the chips wrapper is always rendered, never `display: contents`;
- the trigger precedes chip close buttons in tab order; Backspace removes the last chip in both modes;
- the field click forwarder depends on a DOM ref, never an attribute;
- the native mirror gets unfiltered `@items`;
- every `data-component` / `data-test-id` hook survives unchanged.

The chips-wrapper element is byte-identical to #509 (token-compared), so invariants 3 and 4 hold by construction rather than by re-derivation.

## Verification

880 tests / 877 pass / 3 skip / 0 fail — exactly the #509 baseline. `pnpm build` green, `pnpm lint:js` 0 errors, `frontile lint:types` clean, site builds 68 rendered / 0 failed. The test-app's pre-existing `lint:types` error set is byte-identical to #509's (verified by reverting, rebuilding and diffing).

## Not done

`select.gts` could lose ~250 more lines by extracting the two-way selection state machine (`_selectedKey`/`_selectedKeys`, the sync modifiers, `validateArgs`, the change handlers). It is a coherent seam but touches the most delicate code in the component — the documented backtracking-rerender hazards — so it is left as a separate decision rather than folded in here.

### DIFF
--- a/packages/frontile/src/components/forms/select/native-mirror.gts
+++ b/packages/frontile/src/components/forms/select/native-mirror.gts
@@ -1,0 +1,99 @@
+import Component from '@glimmer/component';
+import {
+  NativeSelect,
+  type ListItem,
+  type NativeSelectSignature
+} from '../native-select';
+
+const eq = (a: unknown, b: unknown) => a === b;
+
+interface SelectNativeMirrorSignature<T> {
+  Args: {
+    /**
+     * The **unfiltered** collection. The mirror carries the form value, which
+     * must not shrink to whatever the listbox filter currently shows, so it is
+     * deliberately never handed `filteredItems`.
+     */
+    items?: T[];
+    allowEmpty?: boolean;
+    disabledKeys?: string[];
+    selectionMode?: 'single' | 'multiple';
+
+    /** Read in multiple selection mode. */
+    selectedKeys: string[];
+    /** Read in single selection mode. */
+    selectedKey: string | null;
+
+    /** Selection handler for multiple selection mode. */
+    onSelectionChange: (keys: string[]) => void;
+    /** Selection handler for single selection mode. */
+    onSingleSelectionChange: (key: string | null) => void;
+
+    onItemsChange: (nodes: ListItem[], action: 'add' | 'remove') => void;
+    placeholder?: string;
+    id?: string;
+    name?: string;
+    isDisabled?: boolean;
+  };
+  Blocks: {
+    default: NativeSelectSignature<T>['Blocks']['default'];
+  };
+}
+
+/**
+ * The visually hidden native `<select>` that makes a Select submit with a form.
+ *
+ * Single and multiple selection mode differ only in three arguments, so those
+ * three are curried onto the component and everything else -- the element, its
+ * attributes and both blocks -- is written **once**. Two copies of this used to
+ * sit side by side in `select.gts`, and a fix to the submitted value landed in
+ * one of them and not the other.
+ */
+class SelectNativeMirror<T = unknown> extends Component<
+  SelectNativeMirrorSignature<T>
+> {
+  <template>
+    {{#let
+      (if
+        (eq @selectionMode "multiple")
+        (component
+          NativeSelect
+          selectionMode="multiple"
+          selectedKeys=@selectedKeys
+          onSelectionChange=@onSelectionChange
+        )
+        (component
+          NativeSelect
+          selectionMode="single"
+          selectedKey=@selectedKey
+          onSelectionChange=@onSingleSelectionChange
+        )
+      )
+      as |Mirror|
+    }}
+      <Mirror
+        @items={{@items}}
+        @allowEmpty={{@allowEmpty}}
+        @disabledKeys={{@disabledKeys}}
+        @onItemsChange={{@onItemsChange}}
+        @placeholder={{@placeholder}}
+        @id={{@id}}
+        @name={{@name}}
+        tabindex="-1"
+        disabled={{@isDisabled}}
+      >
+        <:item as |l|>
+          <l.Item @key={{l.key}}>
+            {{l.label}}
+          </l.Item>
+        </:item>
+        <:default as |l|>
+          {{yield l}}
+        </:default>
+      </Mirror>
+    {{/let}}
+  </template>
+}
+
+export { SelectNativeMirror, type SelectNativeMirrorSignature };
+export default SelectNativeMirror;
--- a/packages/frontile/src/components/forms/select/select.gts
+++ b/packages/frontile/src/components/forms/select/select.gts
@@ -707,8 +707,12 @@ class Select<T = unknown> extends Component<SelectSignature<T>> {
               {{p.measureWidth}}
               {{on "click" this.handleFieldClick}}
               data-test-id={{if this.showChips "chips-field"}}
-              data-has-chips={{this.showChips}}
-              data-invalid={{c.isInvalid}}
+              {{! These flags are stringified rather than handed the boolean.
+              Glimmer renders a true boolean attribute as an empty value, so
+              passing isInvalid straight through produced an empty data-invalid
+              and the theme selector keyed on the value true never matched. }}
+              data-has-chips={{if this.showChips "true" "false"}}
+              data-invalid={{if c.isInvalid "true" "false"}}
               data-disabled={{if @isDisabled "true" "false"}}
               class={{this.classes.chipsField
                 class=@classes.chipsField
--- a/packages/frontile/src/components/forms/select/select.gts
+++ b/packages/frontile/src/components/forms/select/select.gts
@@ -432,14 +432,21 @@ class Select<T = unknown> extends Component<SelectSignature<T>> {
     // with no remembered item would otherwise be counted but never render a
     // chip at all). Every chip's key is in that collection by construction, so
     // the rule's answer is the same for all of them; it is asked of the last
-    // chip, the one Backspace removes. With no chips there is nothing to
-    // remove and nothing that reads this.
+    // chip, the one Backspace removes.
     const keys = this.selectedItems.map((item) => item.key);
     const lastKey = keys[keys.length - 1];
-    return (
-      lastKey !== undefined &&
-      canDeselectKey(keys, lastKey, this.args.allowEmpty)
-    );
+
+    // With nothing selected there is no chip, so nothing reads this today --
+    // but the answer is deliberately `@allowEmpty`, not `false`. That is what
+    // this getter has always returned for an empty selection, and keeping it
+    // means the answer never depends on whether a chip happens to exist right
+    // now. `canDeselectKey` cannot be asked here (it needs a key), so the
+    // empty case is answered directly. Do not "simplify" this away.
+    if (lastKey === undefined) {
+      return this.args.allowEmpty === true;
+    }
+
+    return canDeselectKey(keys, lastKey, this.args.allowEmpty);
   }
 
   /**
--- a/packages/frontile/src/components/forms/select/select.gts
+++ b/packages/frontile/src/components/forms/select/select.gts
@@ -1,463 +1,41 @@
 import Component from '@glimmer/component';
 import { tracked } from '@glimmer/tracking';
-import { on } from '@ember/modifier';
-import { concat, fn } from '@ember/helper';
 import type Owner from '@ember/owner';
-import { NativeSelect, type ListItem } from './native-select';
-import { Listbox, type ListboxSignature } from '../collections/listbox';
-import {
-  useStyles,
-  type SelectSlots,
-  type SelectVariants,
-  type SlotsToClasses
-} from '@frontile/theme';
-import { Spinner } from '../utilities/spinner';
-import { VisuallyHidden } from '../utilities/visually-hidden';
-import { ref } from '../../utils/ref';
-import {
-  Popover,
-  type PopoverSignature,
-  type ContentSignature
-} from '../overlays/popover';
-import { FormControl, type FormControlSharedArgs } from './form-control';
-import { triggerFormInputEvent } from '../../utils/forms-utils-index';
-import { CloseButton } from '../buttons/close-button';
-import { Chip, type ChipSignature } from '../buttons/chip';
-import { IconChevronUpDown } from './icons';
+import { on } from '@ember/modifier';
+import { action } from '@ember/object';
+import { later } from '@ember/runloop';
+import { modifier } from 'ember-modifier';
+import { useStyles } from '@frontile/theme';
+
+import type { ListItem } from '../native-select';
+import { Listbox } from '../../collections/listbox';
+import { VisuallyHidden } from '../../utilities/visually-hidden';
+import { Popover } from '../../overlays/popover';
+import { FormControl } from '../form-control';
+import { ref } from '../../../utils/ref';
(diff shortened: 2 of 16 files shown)


packages/frontile/src/components/forms/select/native-mirror.gts
```diff
@@ -0,0 +1,99 @@
...
+      <Mirror
+        @items={{@items}}
+        @allowEmpty={{@allowEmpty}}
+        @disabledKeys={{@disabledKeys}}
+        @onItemsChange={{@onItemsChange}}
+        @placeholder={{@placeholder}}
+        @id={{@id}}
+        @name={{@name}}
+        tabindex="-1"
+        disabled={{@isDisabled}}
```
josemarluedke, line 83: `disabled={{@isDisabled}}` on the hidden native `<select>` is guarded by nothing. I deleted this attribute, rebuilt, and ran the full suite: **880 tests, 0 failures.**

The failure it would let through: `<Select @isDisabled={{true}} />` renders a disabled trigger but an *enabled* native control inside the `VisuallyHidden` wrapper — keyboard-focusable, changeable, and its value still submitted with the form.

The test that looks like it covers this is `select-test.gts:369`:

```js
assert.dom('[data-component="native-select"]').isDisabled;
```

That is a property access, not a call. `isDisabled` is a qunit-dom *method*; reading it and discarding it asserts nothing. Line 370 (`select-trigger`) has the same bug — that one survives only because the Field and Form `@disabled` tests happen to cover it (removing `disabled` from `trigger.gts` fails exactly 2 tests, both in `field-test`/`form-test`, none in `select-test`).

Not introduced by this PR, but this PR moves the attribute across a new file boundary and cites the unmodified suite as its proof of preservation — so this is precisely the kind of hole the claim rests on. Two `()` in `select-test.gts` would close it.

_🤖 Reviewed by [Claude Code](https://claude.com/claude-code)_

test-app/tests/unit/collections/list-manager-test.ts
```diff
@@ -123,4 +123,136 @@ module('Unit | Utils | ListManager', function (hooks) {
...
+        'one of several always goes'
+      );
+      assert.false(
+        canDeselectKey(['a', 'b'], 'c'),
+        'a key that is not selected has nothing to remove'
+      );
+      assert.false(canDeselectKey([], 'a', true), 'nothing to remove');
+    });
+
+    test('canDeselect reports what selecting the item would do', function (assert) {
```
josemarluedke: This test cannot fail. It asserts

```js
assert.strictEqual(canDeselect, !reported[0]?.includes('a'), ...)
```

but `manager.canDeselect` and `#toggleSelectedItem` both route through the same `canDeselectKey`, so the two sides move together for *any* implementation of the rule.

Verified by mutation — I replaced the whole body of `canDeselectKey` with `return true;`, rebuilt, and ran the module:

```
not ok  canDeselectKey answers for the keys of a selection
ok      canDeselect reports what selecting the item would do   <-- still green
not ok  canDeselect is false for a key that is not selected
ok      canDeselect counts selections whose items are not rendered
```

`return false;` passes it too, by the same symmetry. Six cases across both selection modes and both `allowEmpty` values, and none of them constrains anything the type system doesn't already: it pins self-consistency, which was guaranteed by construction the moment the rule became one function.

The correctness of the extraction is actually pinned by `canDeselectKey answers for the keys of a selection` (which does fail when stubbed) and by `canDeselect counts selections whose items are not rendered`. If you want this test to earn its place, assert the *expected* selection per case (`assert.deepEqual(reported[0], ['b'])` etc.) rather than agreement with the predicate that produced it.

_🤖 Reviewed by [Claude Code](https://claude.com/claude-code)_

packages/frontile/src/utils/listManager.ts
```diff
@@ -402,8 +427,24 @@ class ListManager {
...
-    let selectedKeys: string[] = [];
+  /**
+   * Whether `key` may be removed from the current selection, i.e. whether
+   * clicking an already-selected item deselects it or leaves it selected.
+   *
+   * This is the `allowEmpty` rule as `selectItem` applies it, exposed for
+   * consumers that offer a second route to the same deselection (Select's chip
+   * close buttons) and must agree with the list about whether it is allowed.
+   */
+  canDeselect(key: string): boolean {
```
josemarluedke: `canDeselect` has no production caller. Select's `chipsRemovable` calls the free `canDeselectKey` against its own `selectedItems` array, not the manager (`grep -rn 'canDeselect\b' packages test-app` → this definition, the emitted `.d.ts`, and the unit test).

So the commit message's guarantee — "`ListManager#canDeselect` exposes it for the manager's own selection" so "the dropdown and a chip's close button cannot disagree" — rests entirely on the shared pure function, not on this method. The method is new public API on a public util (it ships in `declarations/utils/listManager.d.ts`) whose only consumer is the test flagged above.

Either wire `chipsRemovable` through the listbox's actual manager, which is what would make the two routes structurally unable to disagree, or drop the method and keep the free function as the single rule.

_🤖 Reviewed by [Claude Code](https://claude.com/claude-code)_

packages/frontile/src/components/forms/select/select.gts
```diff
@@ -836,11 +423,23 @@ class Select<T = unknown> extends Component<SelectSignature<T>> {
...
+    // with no remembered item would otherwise be counted but never render a
+    // chip at all). Every chip's key is in that collection by construction, so
+    // the rule's answer is the same for all of them; it is asked of the last
+    // chip, the one Backspace removes. With no chips there is nothing to
+    // remove and nothing that reads this.
+    const keys = this.selectedItems.map((item) => item.key);
+    const lastKey = keys[keys.length - 1];
+    return (
+      lastKey !== undefined &&
+      canDeselectKey(keys, lastKey, this.args.allowEmpty)
```
josemarluedke: `chipsRemovable` changes its answer at zero selections, which the commit claims it doesn't.

Before: `this.args.allowEmpty === true || this.selectedItems.length > 1` → `true` when `@allowEmpty` is set and nothing is selected.
After: `keys.length === 0` ⇒ `lastKey === undefined` ⇒ `false`.

Concretely: with `@allowEmpty={{true}}` and an empty selection, `removeSelectedKey(key)` used to fall through its guard and call `onSelectionChange([])`, which notifies the parent and dispatches the form `input` event. It now returns early and does neither.

Unreachable from the template today — with no chips there is no close button, and `removeLastSelectedKey` bails on `last`, so nothing reads it in that state. So this is inert rather than a bug. But it is a semantic change inside a commit whose whole thesis is that there aren't any, and the comment above ("With no chips there is nothing to remove and nothing that reads this") documents the reasoning without noting that the old expression answered differently. Worth either restoring the old answer for the empty case or saying explicitly in the comment that the empty-selection answer flipped and why that's safe.

_🤖 Reviewed by [Claude Code](https://claude.com/claude-code)_

josemarluedke (commented):
I reviewed this as a claim to be tested rather than accepted, and **the behaviour-preserving claim holds for the production code**. I re-derived every moved argument, attribute, modifier, block and guard against `origin/main`'s `select.gts`. `test-app/` really is untouched apart from two *added* files — `select-test.gts` is byte-identical. Locally: 880 tests / 877 pass / 3 skip / 0 fail, `frontile build` and `lint:types` clean.

Specifically cleared, since these were the risky seams:

- **The three-arg mirror collapse is correct in both modes.** `selectionMode`, `selectedKeys`↔`selectedKey`, and `onSelectionChange`↔`onSingleSelectionChange` all map to the arm they came from; `@items` (unfiltered), `@onItemsChange`, `@placeholder`, `@id`, `@name`, `tabindex`, `disabled` and both blocks reach the same place. The old `{{#if (has-block "item")}}` inside the mirror's `:item` block had two identical branches, so collapsing it is a genuine no-op.
- **`canDeselectKey` is an exact extraction** of the inline condition (`includes && ((allowEmpty && len==1) || len>1)`), and `#selectionSnapshot` is the unchanged rebuild loop, multiple-only unrendered-key carry-over included. Zero-selection and single-mode included — see the one caveat in the inline comment on `chipsRemovable`.
- **`ListItem.item` introduces no new retention and no new invalidation.** `el` was already held, so the lifetime class is unchanged. And adding `item=@item` to `setupItem`'s args cannot cause extra unregister/register churn, because `{{#each @items}}` keys on `@identity`: if the entry's identity changes the block is rebuilt regardless, and if it doesn't, `item` doesn't change either. The backtracking hazard in `listManager.ts` is untouched. `undefined` for block-form options and for the `allowEmpty` placeholder option, as documented.
- **The mirror does not remount on every selection change**, which was my main worry about currying `selectedKeys` onto `(component NativeSelect …)`. Glimmer's curry opcode caches on the inner definition (`value === lastValue`) and captures args as references, so a new `@selectedKeys` array does not produce a new `CurriedValue` (@glimmer/runtime 0.94.10, opcode 77).
- **No hardcoded Tailwind utility** in any of the six new `.gts` files — every class comes from `useStyles().select(...)` handed down as `@classes`, so nothing lands outside the theme's `@source`.
- `forms/index.ts`'s `export * from './select'` resolves to the new barrel; public exports unchanged (`SelectedItem` added). `signature-data.ts` goes 92 → 97 entries with none lost, no duplicate `name`, and `Select` still at 92 rows — and `select.md` uses explicit `<Signature @component="Select" />`, so the five new internals don't leak onto the docs page.

My findings are all about the **evidence**, not the code. Two of the assertions that are supposed to guard this restructure cannot fail, which matters because the PR's headline argument is "877 unmodified assertions pass".

Two additional coverage gaps that don't warrant their own thread, but are worth knowing given this file's history:

- `data-invalid` and `data-has-chips` on the chips-field wrapper have **zero** assertions anywhere in the suite.
- `endContentPointerEvents` and the `input-end-content` hook in the new `end-content.gts` have zero assertions; only the `div:last-child` position of the `endContent` block is pinned.
- `tabindex="-1"` on the mirror is untested. (`@allowEmpty` and `@disabledKeys` on the mirror *are* — removing them fails 4 tests, so that half of the collapse is real evidence.)

packages/frontile/src/components/forms/select/native-mirror.gts
josemarluedke: Fixed in `3068aed3`. Added the missing call parentheses to both no-op assertions in `select-test.gts:369-370`. Verified green as repaired, then red on deleting `disabled={{@isDisabled}}` from `native-mirror.gts` (`Element [data-component="native-select"] is not disabled`), then restored. A repo-wide grep for the same matcher-as-property mistake across `test-app/tests/` found no others.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

test-app/tests/unit/collections/list-manager-test.ts
josemarluedke: Fixed in `b9d5b3ac`. Retargeted at the real risk — call-site agreement — rather than deleted: it now compares the free `canDeselectKey` against what toggling actually does, across the same six-case matrix, with a comment stating it guards agreement and *not* the rule's correctness (the direct assertions above cover that). Verified red by inverting the condition at the `#toggleSelectedItem` call site only.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/frontile/src/utils/listManager.ts
josemarluedke: Fixed in `b9d5b3ac`. Removed `ListManager#canDeselect` entirely — no reference remains in the repo. `Select#chipsRemovable` and `#toggleSelectedItem` both go through the free `canDeselectKey`, so the rule still lives in one place without the unused public surface.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/frontile/src/components/forms/select/select.gts
josemarluedke: Fixed in `929a34d9`. `chipsRemovable` now answers the zero-selection case explicitly as `this.args.allowEmpty === true`, restoring exact pre-refactor parity for every input, with a comment recording that it is deliberate. The claim is now literally true rather than true-by-reachability-argument.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #514, opened 2026-08-28
fix(forms): guard unflattenData against prototype pollution

josemarluedke:
## Problem

`unflattenData` is vulnerable to prototype pollution.

It splits a dotted path and walks/creates nested objects. For the segment `__proto__`, `isPlainObject(current['__proto__'])` returns `true` — it *is* `Object.prototype` — so the "create a fresh object" branch is skipped, `current` becomes `Object.prototype`, and the final assignment writes onto every object in the application.

Verified against the unmodified function body:

```
unflattenData({'__proto__.isAdmin': true, 'email':'a@b.c'})
  => returns {"email":"a@b.c"}
  => ({}).isAdmin === true      // global pollution
```

The vector is reachable through the public `Form` component: `handleInput`, `handleSubmit`, and `handleReset` all run `unflattenData` over keys taken from the `name` attributes of the form's controls. Apps that render fields from a server-supplied schema, a CMS, or URL state let untrusted input choose those names.

CWE-1321. There was no existing test coverage for this.

## Fix

Refuse `__proto__`, `constructor`, and `prototype` as path segments, anywhere in a path, in both `unflattenData` and `flattenData`. An entry with an unsafe segment is dropped wholesale rather than partially materialized.

`__proto__` is the live vector. `constructor` and `prototype` do **not** reach `Object.prototype` through the walk as written (`isPlainObject` rejects the function they resolve to, so the walk shadows them with a fresh own key) — they are refused as defense in depth, so form data cannot shadow those names and a later change to the walk cannot quietly turn them into live vectors. The code comment says exactly this rather than overstating the risk.

Ordinary data keeps its shape and its normal prototype: the result is consumed by Glimmer templates and Ember's `get()`, so deliberately **not** `Object.create(null)`. A test pins that.

## Behavior change

`flattenData`/`unflattenData` are publicly exported, so these three names are now reserved: a field named e.g. `constructor` is silently dropped from the form data. Kept silent deliberately — these run on every input event — and documented in `form.md`.

## Tests

9 added; 5 of them confirmed failing before the fix.

- `unflattenData` does not pollute via `__proto__` / via `constructor.prototype` / via a dangerous segment mid-path / via a bare `__proto__` final segment
- `unflattenData` returns objects with a normal prototype (guards against a future `Object.create(null)` "fix")
- `flattenData` drops dangerous keys so they cannot round-trip
- Integration: a hostile field name inside a real `<Form>` cannot pollute, and safe fields still come through
- Regression lock-ins: `File` instances pass through both functions by identity

Both pollution submodules scrub `Object.prototype` in `afterEach`, so a regression cannot poison the rest of the suite run.

Full suite: **906 tests, 903 pass, 3 skip, 0 fail** (897 baseline + 9 new). `lint:types` clean.

### DIFF
--- a/packages/frontile/src/components/forms/form.md
+++ b/packages/frontile/src/components/forms/form.md
@@ -822,6 +822,11 @@ object internally and unflattens it again for `@onChange` and `@onSubmit`, so yo
 your `@data` keep the shape you actually want. Errors and dirty entries are keyed by the same
 dotted path down to the leaf — `user.name.first`, never just `user`.
 
+`__proto__`, `constructor` and `prototype` are reserved: a field whose name uses one of those
+segments is dropped from the form data rather than nested. Building those keys out of untrusted
+field names would let a form write onto `Object.prototype` and change every object in the
+application, so they are refused outright.
+
 ```gts preview
 import Component from '@glimmer/component';
 import { tracked } from '@glimmer/tracking';
--- a/packages/frontile/src/utils/nested-data.ts
+++ b/packages/frontile/src/utils/nested-data.ts
@@ -29,6 +29,46 @@ function isPlainObject(value: unknown): value is Record<string, unknown> {
 }
 
 /**
+ * Path segments that are refused as object keys.
+ *
+ * Field names come from the `name` attribute of the form's controls, which an
+ * app may render from a server-supplied schema, a CMS, or URL state — so they
+ * are untrusted input.
+ *
+ * `__proto__` is the real vector: `isPlainObject(Object.prototype)` is `true`,
+ * so when the walk below reached a `__proto__` segment it accepted
+ * `Object.prototype` as an already-existing nested object instead of creating a
+ * fresh one, and the final assignment wrote onto every object in the
+ * application.
+ *
+ * `constructor` and `prototype` do not reach `Object.prototype` through the
+ * walk as it is written — `current['constructor']` is a function, which
+ * `isPlainObject` rejects, so the walk shadows it with a fresh own key. They
+ * are refused anyway, as defense in depth: it keeps form data from shadowing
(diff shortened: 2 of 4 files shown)


josemarluedke:

**Spec axis: conforms, no findings.** Standards axis: no documented-standard breaches; three judgement calls, none blocking.

Worth recording what the reviewers checked rather than assumed:

- The guard blocks unsafe segments at **any** depth, matching *"not just first"*.
- The normal-prototype requirement is **pinned by a test** (`returns objects with a normal prototype`), which is what a future `Object.create(null)` "simplification" would break for Glimmer and Ember's `get()`.
- The unit tests build the own `__proto__` key via `Object.defineProperty` — the only way to make the input reach the buggy branch. A plain object literal would have made them vacuous, and they avoid it.
- The integration test asserts `({}).isAdmin === undefined` after both `fillIn` **and** submit, and fails pre-fix.
- `Object.prototype` scrubbing in `afterEach` prevents cross-test contamination.

- **The `constructor`/`prototype` cost is real and worth a conscious sign-off.** The JSDoc states outright that two of the three keys guard a need that doesn't exist on the code as written. The consequence: a legitimate field named `constructor` is now silently dropped from form data. I still think defence-in-depth is right here, and the comment is now honest about it — but it's a public behaviour change, so it should be a deliberate decision rather than one inherited from a comment.
- **`isSafePath` is a one-line wrapper with a single call site.** `if (keys.some(isUnsafeKey)) continue;` inlined loses nothing; the doc comment above it carries the actual explanation.
- The two `File` pass-through tests were read as unrelated to a security fix. They aren't scope creep — I asked for pass-through regression coverage explicitly — but they are near-identical in shape and could share a helper.

PR #515, opened 2026-08-28
fix(press): stop swallowing document-level pointerup

josemarluedke:
## Problem

Two defects in the `press` modifier's release handling.

**A. Every `pointerup`/`mouseup` in the document was swallowed.** `addDocumentListeners` registered `end` on `document` in the **capture** phase, and `end` called `event.stopPropagation()` by default. Document capture is the first node in the propagation path, so after any press the release event never reached *any* element below it.

Failure: an app with a `mouseup` listener for drag-end, a canvas interaction, a resizer, or "dismiss on mouseup outside" stops receiving that event as soon as the user presses any `Button`, `ToggleButton`, or `CloseButton`. Even the pressed element's own `pointerup`/`mouseup` listeners were dead.

The documented intent of the default `stopPropagation` is to keep a press from bubbling to *parent elements* — not to kill the event for the whole page.

**B. `onPressUp` fired for releases outside the element.** It was constructed and invoked with no containment check, unlike `onPress`. Press a button, drag off it, release: `onPress` correctly skipped, `onPressUp` fired anyway. Contrary to the react-aria semantics this API mirrors.

## Fix

The document listeners are now a true outside-only fallback: `endFromDocument`/`cancelFromDocument` return immediately when `element.contains(event.target)`, letting the element's own bubble-phase listener handle that case. `stopPropagation()` is only called for a release that actually targeted the element.

They stay in the **capture** phase deliberately — that is what makes the outside-release fallback robust (an intermediate handler that stops propagation can't strand the modifier in a permanently-pressed state). Deferring rather than re-phasing keeps that property while restoring the documented intent.

`onPressUp` is now gated on containment, consistent with `onPress`.

Three follow-on changes were forced by the above and are the parts most worth reviewing:

1. **`continuePropagation()` now works.** A single DOM release produces up to three `PressEvent`s over one shared underlying event, and each previously stopped propagation independently — so a consumer who provided only `onPress` and called `continuePropagation()` there was overruled by the `pressend` event, which is constructed even with no `onPressEnd` callback. Propagation now continues if *any* handler asked for it. This only makes a previously non-functional API work, but it is a semantic choice (OR rather than AND) and is now documented in `press.md`.
2. **`pointercancel` is bound on the element whenever end events are tracked**, not only when `onPressEnd` was given. With the document fallback ignoring element-targeted events, an `onPressChange`-only consumer would otherwise have lost cancel handling entirely — caught by the pre-existing `onPressChange called on pointer cancel` test.
3. **`cancel()`'s `stopPropagation()` is now conditional too** (`isInsideElement(event) && !pressEvent.shouldPropagate`). `cancel` is reached from both the element listener and the document fallback, so it needed the same containment + opt-in gate as `end` — otherwise a `pointercancel` originating outside the element would still be swallowed, and `continuePropagation()` would be honoured on release but ignored on cancel.

## Tests

5 added, 4 confirmed failing before the fix (`# tests 57 / # pass 53 / # fail 4` on the filtered run):

- A release outside the element is not swallowed for other listeners (document `pointerup` **and** `mouseup`)
- The pressed element's own `pointerup`/`mouseup` listeners still fire
- `onPressUp` does not fire when released outside the element, while `onPressEnd` still does and `onPress` still doesn't
- Press events bubble to ancestor handlers when `continuePropagation()` is called
- Press events do **not** bubble to ancestors by default — added deliberately as the guard that this fix doesn't over-correct

Document-level listeners added by tests are removed in a `finally` so they cannot leak into later tests.

**One pre-existing test was changed:** `drag outside element and release still triggers end events` asserted 3 captured events including an `up` — it encoded defect B. It now expects 2 (`start`, `end`) plus an explicit assertion that no `up` event is captured. No assertions were removed; the `onPress`-not-fired assertion is intact.

Full suite: **902 tests, 899 pass, 3 skip, 0 fail** (897 baseline + 5 new). `--filter="press"` 57/57. `lint:types` clean.

## Verification status

Verified through the integration suite, which runs in real Chrome. Interactive browser verification of the downstream consumers (Button, CloseButton, Chip, and overlay outside-click on a live page) is being done centrally on an integration branch of all the current fixes; I'll follow up here if it turns anything up.

### DIFF
--- a/packages/frontile/src/modifiers/press.md
+++ b/packages/frontile/src/modifiers/press.md
@@ -81,13 +81,13 @@ export default class PressExample extends Component {
 
 The press modifier accepts an optional positional `onPress` function and the following named options:
 
-| Name            | Type                           | Description                                                                                                     |
-| --------------- | ------------------------------ | --------------------------------------------------------------------------------------------------------------- |
-| `onPress`       | `(e: PressEvent) => void`      | Handler called when press is released over the target.                                                          |
-| `onPressStart`  | `(e: PressEvent) => void`      | Handler called when a press interaction starts.                                                                 |
-| `onPressEnd`    | `(e: PressEvent) => void`      | Handler called when a press interaction ends, either over the target or when the pointer leaves the target.     |
-| `onPressUp`     | `(e: PressEvent) => void`      | Handler called when a press is released over the target, regardless of whether it started on the target or not. |
-| `onPressChange` | `(isPressed: boolean) => void` | Handler called when the press state changes.                                                                    |
+| Name            | Type                           | Description                                                                                                         |
+| --------------- | ------------------------------ | ------------------------------------------------------------------------------------------------------------------- |
+| `onPress`       | `(e: PressEvent) => void`      | Handler called when press is released over the target.                                                              |
+| `onPressStart`  | `(e: PressEvent) => void`      | Handler called when a press interaction starts.                                                                     |
+| `onPressEnd`    | `(e: PressEvent) => void`      | Handler called when a press interaction ends, either over the target or when the pointer leaves the target.         |
+| `onPressUp`     | `(e: PressEvent) => void`      | Handler called when a press is released over the target. Not called when the release happens outside of the target. |
+| `onPressChange` | `(isPressed: boolean) => void` | Handler called when the press state changes.                                                                        |
 
 ### PressEvent
 
@@ -114,3 +114,8 @@ Press events are normalized across different input methods and provide consisten
 - **Performance**: Only adds event listeners for callbacks you provide
 - **Accessibility**: Automatically handles Enter and Space key interactions
 - **Cross-platform**: Supports mouse, touch, and keyboard across all browsers
+- **Propagation**: By default a press stops propagation of the underlying
+  event to parent elements. Call `continuePropagation()` on the `PressEvent` in
+  any handler to let it through. Only events on the pressed element itself are
+  stopped - a release that happens outside of the element is left untouched, so
+  other listeners on the page (including document-level ones) still receive it.
--- a/packages/frontile/src/modifiers/press.md
+++ b/packages/frontile/src/modifiers/press.md
@@ -111,7 +111,9 @@ Press events are normalized across different input methods and provide consisten
 - **Positional argument**: Pass `onPress` as the first argument: `{{press this.handlePress}}`
 - **Mixed arguments**: Combine positional `onPress` with named arguments for other callbacks
 - **Conflict handling**: Cannot provide both positional and named `onPress` (assertion error)
-- **Performance**: Only adds event listeners for callbacks you provide
+- **Performance**: Only adds the event listeners your callbacks require - a
+  single callback can need more than one (e.g. `onPressChange` also needs
+  `pointercancel`)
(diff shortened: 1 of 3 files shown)


josemarluedke:

**Both axes essentially clean.** Spec axis confirmed both findings implemented, only the two accepted deviations present (`continuePropagation()` OR semantics, `pointercancel` widened to `needsEndEvents`), and `shouldIgnoreEvent` dedup plus the keyboard path intact.

Test evidence verified as genuinely discriminating: `stopPropagation()` in document capture halts the whole path *including* the document's own bubble phase, so both document-listener counters and the element's own `pointerup`/`mouseup` counters fail on `main`. The `onPressUp`-outside test asserts the tri-state (`end` 1, `press` 0, `up` 0), which is Finding B exactly. Document listeners are removed in `finally`, so no cross-test leakage.

### 🟡 Should fix — a public doc bullet is now slightly false

`press.md` still says *"**Performance**: Only adds event listeners for callbacks you provide"*. Widening `if (named.onPressEnd)` to `if (needsEndEvents)` means a consumer passing only `onPressChange` now gets an element-level `pointercancel` listener. The change is correct and the code comment explains it — but that file renders as the public doc, so the bullet needs a qualifier.

- **`isOnTarget` is a mysterious name in this file specifically.** It means "inside the *pressed element*", but "target" already means `event.target` here, and the function's whole job is to distinguish the two. `isInsideElement` or `isOwnEvent` reads truer.
- **Duplicated wrappers:** `endFromDocument` and `cancelFromDocument` are the same shape. One `onlyOutsideTarget(handler)` factory would express the "document listeners are a fallback" rule once and keep the add/remove listener pairs symmetric by construction.
- **Accumulator shape:** `let shouldPropagate = false` is declared before the first event exists, then folded three times. Collecting the events and asking `events.some((e) => e.shouldPropagate)` once at the stop site would state the rule ("any handler opting in wins") rather than accumulating it.
- One change isn't named in the finding list or the PR body's "two follow-on changes": `cancel()`'s `stopPropagation()` also became conditional. It falls inside the letter of the brief ("or restructure so the document fallback never stops propagation") and is required for consistency — worth a line in the body for completeness.
- `pressstart` still stops propagation on its own flag rather than the OR'd accumulator. Correct: the OR is per-release, and `pressstart` has a different underlying event.

josemarluedke:
Review feedback addressed in d5efc49e.

**Doc accuracy:** the `press.md` performance bullet no longer implies one listener per callback — it now reads "Only adds the event listeners your callbacks require - a single callback can need more than one (e.g. `onPressChange` also needs `pointercancel`)".

**PR body:** the follow-on list is now three items; added `cancel()`'s conditional `stopPropagation()` with its rationale (same containment + opt-in gate as `end`, since `cancel` is reached from both the element listener and the document fallback).

**Optional suggestions — took all three:** `isOnTarget` → `isInsideElement`; an `onlyOutsideElement` factory replacing the duplicated document wrappers (declared after `end`/`cancel` to avoid a TDZ, identities stored in consts so the add/remove pairs stay symmetric); and `dispatched.some((e) => e.shouldPropagate)` in place of the folded accumulator.

902 tests / 0 fail, `--filter="press"` 57/57, `lint:types` clean.

PR #516, opened 2026-08-28
fix(collapsible): ignore transitionend from descendants and stop leaking waiters

josemarluedke:
## Problems

**1. A descendant's `transitionend` mutated consumer DOM.** `transitionend` bubbles, and `onTransitionEnd` never checked the event target before doing `(event.target as HTMLElement).style.height = 'auto'` and clearing `overflow`.

So any child inside an open collapsible that transitions `height` or `opacity` — a hover card, a nested accordion, a fading image, a Tailwind `transition-opacity` button — got `height: auto` stamped onto it as an inline style, permanently overriding its own layout.

**2. The test waiter was leaked and double-ended.** One `waiterToken` field, two defects:
- Toggling twice before the first transition finished called `beginAsync()` again and overwrote the token. The first was never ended, so `settled()` / `await click()` never resolves — a hang in the *consumer's* test suite.
- When opening, both the `height` branch and the `opacity == '1'` branch can match, so `endAsync` ran twice with the same token (never nulled).

**3. `@initialHeight` could inject extra CSS declarations.** It reaches an `htmlSafe` style string via `safeStyles`, which interpolates values verbatim, so `@initialHeight="10px; position: fixed"` smuggled in a second declaration. Not an XSS vector (Glimmer sets attributes through the DOM API), but a consumer-controlled value should not be able to add declarations.

## Fixes

- Ignore `transitionend` whose `target` is not the collapsible itself.
- Route every end through `endWaiter()`, which clears `waiterToken` *before* ending it so the same token can never be ended twice; end any in-flight token before `beginAsync()`; end the token in `willDestroy` so a teardown mid-transition doesn't leave a waiter pending.
- Validate `@initialHeight` in `Collapsible` itself; `safeStyles` is left exactly as it was on `main`. A shared helper that silently rewrites its caller's values is the wrong place for this, so the check lives where the untrusted value enters. Rather than a length regex — easy to get wrong on `calc`/`min`/`clamp`/`var()` — the value is assigned to a detached element's `style.height` and read back, letting the browser's own CSS parser decide: it keeps every real height and drops anything it cannot parse as one, a trailing declaration included. A rejected value falls back to the collapsed default of `0`.

## Tests

6 added (4 → 10). Four were confirmed failing before the fix; the two marked below are baseline guards that passed before and after.

- `does not modify descendants when a transitionend bubbles up from a child` — dispatches a synthetic bubbling `transitionend` from a child, asserts the child's inline `height`/`overflow` stay empty, then dispatches on the collapsible itself and asserts its own reset still works
- `does not leak a test waiter when toggled twice before the transition finishes` — polls `getPendingWaiterState()` rather than `settled()`, since a genuinely leaked waiter would hang the test itself
- `ends the test waiter only once per toggle`
- `does not let @initialHeight inject extra CSS declarations`
- `keeps every legitimate @initialHeight value` — `10px`, `2rem`, `50%`, `0`, `calc(1rem + 2px)`, `min()`, `clamp()`, `var()`; **passed before and after**, it exists to prove the new validation is not over-eager
- `resets its own height and overflow when the open transition ends` — the missing baseline coverage; **passed before and after**, added as a guard

No pre-existing test asserted the buggy behavior; the original 4 are untouched.

Full suite: **903 tests, 900 pass, 3 skip, 0 fail** (897 baseline + 6 new). `lint:types` clean.

## Two notes for the reviewer

1. **Defect 2b does not throw in the installed `@ember/test-waiters` 4.1.2** — after a first end the token is in the completed-operations map, so a double-end is silently tolerated rather than asserting. It is still wrong and is fixed, but the failure mode was quieter than expected.
2. **Pre-existing, deliberately not fixed here:** the open animation truncates. The opacity transition finishes at 0.3s and the handler's `height: auto` is not animatable, so the in-flight 0.4s height transition emits `transitioncancel` rather than `transitionend`. A consequence worth knowing: the `isOpen && propertyName === 'height'` branch in `onTransitionEnd` is largely dead in practice when opening without an `@initialHeight` — the opacity branch is what actually ends the waiter. Any future fix for the truncation should re-check that branch logic alongside it. This behavior is unchanged by this PR.

### DIFF
--- a/packages/frontile/src/components/utilities/collapsible.gts
+++ b/packages/frontile/src/components/utilities/collapsible.gts
@@ -7,6 +7,39 @@ import type Owner from '@ember/owner';
 
 const waiter = buildWaiter('frontile:collapsible');
 
+/**
+ * Runs a caller-supplied height through the browser's own CSS parser.
+ *
+ * `@initialHeight` ends up in an inline `style` attribute, so a value such as
+ * `10px; position: fixed` would otherwise smuggle in a second declaration.
+ * Assigning it to a detached element's `style.height` lets the CSSOM decide:
+ * it accepts every real height (`10px`, `2rem`, `50%`, `0`,
+ * `calc(1rem + 2px)`, `var(--x)`, …) and drops anything it cannot parse as a
+ * single height value — a trailing declaration included. Returns the parsed
+ * value, or `undefined` when the input is unusable.
+ */
+function parseHeight(value: string | undefined): string | undefined {
+  if (!value) {
+    return undefined;
+  }
+
+  // Reject anything that could terminate the declaration up front. The CSSOM
+  // check below catches this too, but the explicit test keeps the guarantee
+  // even where `document` has no real CSS parser (SSR / prerender).
+  if (/[;}]/.test(value)) {
+    return undefined;
+  }
+
+  if (typeof document === 'undefined') {
+    return value;
+  }
+
+  const probe = document.createElement('div');
+  probe.style.height = value;
+
+  return probe.style.height || undefined;
+}
+
 interface CollapsibleArgs {
   /**
    * If true, the content will be visible
@@ -15,7 +48,9 @@ interface CollapsibleArgs {
 
   /**
    * The height for the content in it's collapsed state.
-   * The unit of the value should be included, eg. '10px'.
+   * The unit of the value should be included, eg. '10px'. Any CSS height is
+   * accepted (`2rem`, `50%`, `calc(1rem + 2px)`, …); a value the CSS parser
+   * rejects is ignored and the content collapses to `0`.
    *
    * @defaultValue 0
    */
@@ -44,18 +79,28 @@ class Collapsible extends Component<CollapsibleSignature> {
     }
   }
 
+  /**
+   * `@initialHeight`, validated by the CSS parser. `undefined` when it was not
+   * given or could not be parsed as a height, in which case the collapsed
+   * state falls back to `0`.
+   */
+  get initialHeight(): string | undefined {
+    return parseHeight(this.args.initialHeight);
+  }
+
   get styles(): ReturnType<typeof safeStyles> {
     let styles: Record<string, string | number> = {};
+    const initialHeight = this.initialHeight;
 
     if (!this.isInitiallyOpen) {
       styles = {
         ...styles,
-        height: this.args.initialHeight || 0,
-        opacity: this.args.initialHeight ? '1' : '0'
+        height: initialHeight || 0,
+        opacity: initialHeight ? '1' : '0'
       };
     }
 
-    if (this.args.initialHeight || !this.isInitiallyOpen) {
+    if (initialHeight || !this.isInitiallyOpen) {
       styles = {
         ...styles,
         overflow: 'hidden'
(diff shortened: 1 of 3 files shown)


josemarluedke:

**Spec axis: conforms** — all three findings implemented, nothing outside the brief, and the two things I most doubted both hold up:

- The component's own reset survives the new guard: it's `event.target !== event.currentTarget`, and the handler is bound with `{{on "transitionend"}}` on the collapsible's own root div, so its own `transitionend` still reaches the `height: 'auto'` / `overflow: ''` branch. Asserted directly in the second half of the descendant test plus a dedicated test.
- **The waiter tests are not vacuous.** `getPendingWaiterState().waiters` keys a waiter only when `waitUntil()` is false, i.e. only while a token is genuinely outstanding — so pre-fix, the double toggle overwrites token A, only B is ended, and A shows up as pending. Polling-then-asserting rather than calling `settled()` turns that into a failed assertion instead of a hang, which is the only way to observe it.
- `10px`, `2rem`, `50%`, `0`, `calc(1rem + 2px)` all pass through `safeStyles` byte-identical.

### 🟡 Should fix — `safeStyles` silently rewrites caller data, and the name doesn't say so

Flagged independently by both a standards reviewer and me on the first pass. `packages/frontile/src/utils/safe-styles.ts:11-22`:

```js
const value = String(style[key]).replace(/;/g, '');
```

Nothing in the name or the signature (`Record<string, string | number>` → `SafeString`) suggests values get edited, and the added comment **concedes** the failure mode (a `url(data:…;base64,…)` value "would silently corrupt") without preventing it. This is a shared helper re-exported to consumers via `packages/utilities/src/utils/safe-styles.ts`, so the mangling reaches code outside this repo — for the benefit of the single in-repo caller.

**Two better shapes:**
1. Reject-and-warn on a value containing `;` rather than silently stripping; or
2. Push the sanitising to the one caller — `collapsible.gts` only ever needs a length for `height`, so validate `@initialHeight` there (or set it through the CSSOM, as the comment itself recommends), and leave the shared helper honest.

I'd take (2): it keeps the fix and removes the shared-helper blast radius entirely.

- With the new guard at the top of `onTransitionEnd`, the two `(event.target as HTMLElement)` casts below are redundant narrowing of a known element — `event.currentTarget` reads truer.
- `collapsible-test.gts` repeats `document.querySelector('[data-test-id=collapsible]') as HTMLElement` in three of four new tests, and the rAF-flush idiom twice back-to-back in the last one. Both want a local helper alongside the `collapsibleWaiter()` helpers already at the top of the file.
- Minor PR-body inaccuracy: it claims all 5 tests were "confirmed failing before the fix" and then says `resets its own height and overflow` "passed before and after". The latter is correct and was requested as baseline coverage; only the blanket claim is wrong.

josemarluedke:
Addressed the review.

**`safeStyles` reverted to `main`** — no stripping, no `String()` coercion. A shared helper re-exported to consumers shouldn't silently rewrite caller data, and it was doing so for one in-repo caller.

**The sanitising moved to `Collapsible`**, where the untrusted value enters. Instead of a length regex (the `calc`/`min`/`clamp`/`var()` trap), `@initialHeight` is assigned to a detached element's `style.height` and read back — the browser's own CSS parser decides. `contract()` routes through the same getter, so a rejected value collapses to `0` rather than leaving the element expanded.

I reverted the helper first and confirmed `does not let @initialHeight inject extra CSS declarations` **failed**, then re-confirmed it passes with the caller-side check — the protection moved, it didn't vanish.

**New test `keeps every legitimate @initialHeight value`** proves `10px`, `2rem`, `50%`, `0`, `calc(1rem + 2px)`, `min(10px, 2rem)`, `clamp(1px, 2rem, 3rem)` and `var(--x, 4px)` all survive. Worth noting it caught something a regex never would: Chrome re-serialises `calc(1rem + 2px)` as `calc(2px + 1rem)`, so the math functions are matched loosely.

**Also**: `onTransitionEnd` reads `event.currentTarget` once instead of three `event.target` casts; the test file gained `collapsibleElement()` and `flushAnimationFrames()` helpers.

Fixed the PR body's blanket "all confirmed failing" claim — 4 of 6 were; the two baseline guards are now marked.

Full suite: **903 tests, 900 pass, 3 skip, 0 fail**. `lint:types` clean, `lint:js` 0 errors.

PR #517, opened 2026-08-28
fix(status): clamp progress bar percentage and fix value formatting

josemarluedke:
## Problems

**1. `percentage` was unclamped and undefended against a zero range.**
- `@minValue` equal to `@maxValue` → division by zero → `Infinity`/`NaN` → `style="width: NaN%"`, which the CSS parser drops, so the bar silently rendered at zero with no warning.
- `@progress={{150}}` on the default 0–100 scale → 150% width, a bar overflowing its track.
- Negative values → a negative width, also dropped.

**2. `get progress()` swallowed a legitimate `0`.** `this.args.progress || this.minValue` meant `@progress={{0}}` was computed as though progress were `minValue` — the wrong position on the scale and a wrong label. On a `-10`–`10` scale, a real `0` is the midpoint (50%) but rendered at the bottom (0%).

**3. Percent formatting was wrong by 100×.** The arg docs say values are formatted as a percentage by default, but `formattedValueLabel` passed the raw `progress` to `Intl.NumberFormat`, which multiplies a `percent` style by 100 itself. So the documented `@formatOptions={{hash style="percent"}}` with `@progress={{50}}` announced **"5,000%"**.

**4. `navigator?.language` was not SSR-safe.** Optional chaining does not protect an *undeclared* global — if `navigator` doesn't exist this throws `ReferenceError`, and this repo prerenders its documentation site.

## Fixes

- `percentage` clamps to 0–100 and returns 0 when the range is not positive.
- `progress`, `minValue`, and `maxValue` use `??` instead of `||`. (`@maxValue={{0}}` previously became `100` silently — same class of defect; now a zero range, which `percentage` handles safely.)
- `formattedValueLabel` feeds a `percent` style the fraction of the min–max range, and every other style (`decimal`, `currency`, `unit`) the raw `progress`. This makes `style="percent"` a locale-aware equivalent of the default label — the two now agree — and makes it respect `@minValue`/`@maxValue`, so `@progress={{20}}` on a 10–30 scale reads `50%`. Raw-value styles keep the consumer's own units, which is what the existing currency tests expect.
- `typeof navigator !== 'undefined'` guard, following the precedent already documented for `File`/`FileList` in `utils/nested-data.ts`.
- **`aria-valuenow` clamps to the scale too.** ARIA requires it to sit within `valuemin`/`valuemax`, and assistive technology computes the announced percentage from those three. Reporting a raw `150` meant a sighted user saw a bar pinned at 100% while a screen-reader user was told "150%" — the wrong answer going to exactly the people who cannot see the bar. The indeterminate early-return is preserved, so `aria-valuenow` is still *absent* rather than clamped in that state.

Clamping is a display safeguard, not validation: nothing warns, deliberately, since a legitimately jittery value (byte counts overshooting a total) would spam a warning on every render. Documented so nobody adds one later.

## Tests

10 added (22 → 32); the bug reproductions were confirmed failing first, with the exact wrong values in the output — `5,000%` and `2,000%` for the percent style, `width: NaN%` / overflowing widths for the clamp cases, and `width: 0%` where a real `0` should sit mid-scale for the `||` case.

- clamps above `maxValue` / below `minValue` / negative → asserts both the width and the reported `aria-valuenow`
- renders a valid width when `minValue` equals `maxValue` — asserts the `style` contains no `NaN` *and* that `element.style.width === '0%'`, i.e. the CSS parser actually accepted it
- treats a progress of `0` as `0`, not as `minValue` — on a **-10–10 scale that straddles zero**, so the two readings are distinguishable: a real `0` is the midpoint (`width: 50%`, label `50%`, `aria-valuenow="0"`), while the `||` substitution of `minValue` pins it at the bottom (`width: 0%`, label `0%`, `aria-valuenow="-10"`). An earlier version of this test used a 10–30 scale, where `0` and the substituted `10` both compute to `0%` and both clamp `aria-valuenow` to `10` — every assertion was insensitive to the fix. Re-verified by reverting the `??` and watching it fail.
- formats a percent style without scaling it twice, and against the min/max range
- regression guards: indeterminate 50% width (previously unasserted), `@showValueLabel` defaulting, no label row when no label

No assertions were deleted. Full suite: **907 tests, 904 pass, 3 skip, 0 fail** (897 baseline + 10 new). `lint:types` clean.

## Notes

- The default label is derived from the clamped percentage, so `@progress={{150}}` displays `100%` — consistent with both the bar and the now-clamped `aria-valuenow`.
- An inverted scale (`@maxValue` < `@minValue`) resolves to `maxValue` for the reported value and `0%` for the width. Coherent but untested — a caller error with no defined correct rendering.
- The SSR guard is not covered by an executing test: the browser test environment always has `navigator`, so there is no honest way to exercise it from the integration suite. It follows an established in-repo precedent instead.

### DIFF
--- a/packages/frontile/src/components/status/progress-bar.gts
+++ b/packages/frontile/src/components/status/progress-bar.gts
@@ -73,8 +73,11 @@ interface ProgressBarSignature {
     showValueLabel?: boolean;
 
     /**
-     * The display format of the value.
-     * Values are formatted as a percentage by default.
+     * The display format of the value, passed to `Intl.NumberFormat`.
+     *
+     * A `percent` style formats the position on the min/max scale as a
+     * fraction; every other style formats the raw `progress` value. With no
+     * format options, the value is displayed as a whole percentage.
      */
     formatOptions?: Intl.NumberFormatOptions;
 
@@ -116,21 +119,37 @@ class ProgressBar extends Component<ProgressBarSignature> {
   }
 
   get minValue(): number {
-    return this.args.minValue || 0;
+    return this.args.minValue ?? 0;
   }
 
   get maxValue(): number {
-    return this.args.maxValue || 100;
+    return this.args.maxValue ?? 100;
   }
 
+  /**
+   * `@progress={{0}}` is a legitimate value, so this falls back to `minValue`
+   * only when nothing was passed — a truthiness check would report the bar as
+   * already at the bottom of the scale instead of at zero.
+   */
   get progress(): number {
-    return this.args.progress || this.minValue;
+    return this.args.progress ?? this.minValue;
   }
 
+  /**
+   * The position on the scale as 0–100. Clamped, because the result drives an
+   * inline `width` — an out-of-range or `NaN` percentage is dropped by the CSS
+   * parser, which renders as a silently empty bar rather than as an error.
+   * A non-positive range has no meaningful position, so it reports 0.
+   */
   get percentage(): number {
-    return (
-      ((this.progress - this.minValue) / (this.maxValue - this.minValue)) * 100
-    );
+    const range = this.maxValue - this.minValue;
+    if (range <= 0) {
+      return 0;
+    }
+
+    const percentage = ((this.progress - this.minValue) / range) * 100;
+
+    return Math.min(100, Math.max(0, percentage));
   }
 
   get showValueLabel(): boolean {
@@ -158,10 +177,20 @@ class ProgressBar extends Component<ProgressBarSignature> {
         ...(this.args.formatOptions || {})
       };
 
-      return new Intl.NumberFormat(
-        navigator?.language || 'en-US',
-        options
-      ).format(this.progress);
+      // `Intl` scales a `percent` style by 100 itself, so it has to be handed
+      // the fraction of the min–max range — passing the already-scaled
+      // percentage announces `@progress={{50}}` as "5,000%". Every other style
+      // (`decimal`, `currency`, `unit`) formats the raw value on its own scale.
+      const value =
+        options.style === 'percent' ? this.percentage / 100 : this.progress;
+
+      // `navigator` is a browser global with no counterpart in Node, so a bare
+      // `navigator?.language` is a hard `ReferenceError` rather than
+      // `undefined` when a ProgressBar is server-rendered.
+      const locale =
+        typeof navigator !== 'undefined' ? navigator.language : undefined;
+
+      return new Intl.NumberFormat(locale || 'en-US', options).format(value);
     }
 
     return this.percentage.toFixed(0) + '%';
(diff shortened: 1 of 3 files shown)


josemarluedke:

Standards axis: **clean** — the live-demo fences are self-sufficient, `@formatOptions`' JSDoc moves with the behaviour, docs updated alongside. Spec axis verified all four findings plus every accepted decision (`aria-valuenow` clamps; no dev warning; percent style fed the range fraction; `typeof navigator` guard).

But one test doesn't prove what it claims, and **that is my fault, not the author's.**

### 🔴 Must fix — Finding 2's test is vacuous

`it treats a progress of 0 as 0, not as minValue` (`progress-bar-test.gts:320-347`) passes identically with the buggy `||`.

Trace it pre-fix with `@progress={{0}} @minValue={{10}}`: `progress` resolves to `0 || 10` → `10`, so `percentage = ((10-10)/(100-10))*100 = 0`. The asserted `width: 0%` and `0%` label are **what the bug produced**. And the asserted `aria-valuenow="10"` is the bug's raw value *and* the fix's clamped value. All three assertions are insensitive to the fix.

How it got here: the test genuinely did fail first, when the implementation reported a *raw* `aria-valuenow` (`0` vs the bug's `10`). I then required `aria-valuenow` to clamp to the range — correctly, for the ARIA reasons in the PR body — and instructed the assertion be changed to `"10"`. That removed the only observable separating 0 from 10, and neither of us noticed.

The `??` fix itself is right. It just needs a test that can fail.

**Fix:** replace with an observable that separates the two, e.g.
- `@progress={{0}} @minValue={{-10}} @maxValue={{10}}` → width `50%` with the fix, `0%` with the bug; or
- `@formatOptions={{hash style="decimal"}}` → label `0` vs `10`.

Also correct the now-false claims in the in-test comment ("giving a non-zero width") and in the PR body ("confirmed failing first … `aria-valuenow="10"`").

### ⚪ Note (no action requested)
The clamp idiom appears twice against different bounds — `Math.min(100, Math.max(0, percentage))` and `Math.min(maxValue, Math.max(minValue, progress))`. Marginal at two instances, but a named `clamp(value, min, max)` would make the "same rule, different scale" relationship explicit instead of inferred from comments.

josemarluedke:
Fixed the vacuous test. You were right that it proved nothing: on the 10–30 scale, `@progress={{0}}` and the buggy `||` substitution of `10` both compute `width: 0%`, both label `0%`, and both clamp `aria-valuenow` to `10` — so all three assertions were insensitive to the `??` fix.

**Observable chosen:** a `-10`–`10` scale that straddles zero. A real `0` is the midpoint (`width: 50%`, label `50%`, `aria-valuenow="0"`); the substituted `-10` pins to the bottom (`0%`, `-10`). All three assertions now discriminate, so the coverage of "a legitimate `0` must not be swallowed" is kept, not weakened.

**Proof** — reverted `get progress()` to `||`, rebuilt:

```
not ok 18 - ProgressBar > progress value: it treats a progress of 0 as 0, not as minValue
    actual:   width: 0%
    expected: width: 50%
    message:  a real 0 sits at the midpoint of a -10–10 scale, not at the bottom
```

Fix restored; **907 tests, 904 pass, 3 skip, 0 fail**. `lint:types` clean.

Also corrected two false claims: the in-test comment said the bug gave "a non-zero width" (it gave `0%`), and the PR body cited `aria-valuenow="10"` as a pre-fix failure signal — post-clamp that is the *expected* value, so it never was one.

PR #518, opened 2026-08-28
fix(table): stop hijacking keys from interactive cell content

josemarluedke:
## Problems

**1. The row keydown handler hijacked keys from interactive cell content.** `handleRowKeydown` is installed on the `<tr>` and read `event.target` without checking it was the row.

- A button or link in a cell: Tab to it, press Enter or Space → `preventDefault()` ran and the **row's** selection toggled instead of the control activating.
- A text input in a cell: ArrowUp/ArrowDown were `preventDefault()`ed so the caret couldn't move, and because `allRows.indexOf(currentElement)` is `-1` for a non-row target, focus **jumped to the first row of the table**.

**2. Every row was a tab stop.** `tabindex={{if this.hasSelection "0" "-1"}}` — a 200-row selectable table added 200 tab stops. The WAI-ARIA grid pattern requires exactly one tabbable element per composite widget.

## Fixes

- `handleRowKeydown` returns early unless `event.target === event.currentTarget`, so anything bubbling from cell content belongs to that control. The guard sits above both the arrow-navigation and the selection branches.
- Roving tabindex: `tabindex="0"` goes to **the first selected row, falling back to the first row** when nothing is selected, and arrow navigation carries it along with focus. Tabbing back into a table should land you where your work is, not restart at the top. It's derived from `sortedItems`, so it respects sort order rather than raw `@items` order. `handleRowKeydown` also claims the tab stop on a Space/Enter selection, so an external selection change can't yank it away from the row the user is standing on. `data-selectable="true"` and the arrow-navigation query are untouched.

## One reported finding was not real

I had also flagged `getKey`'s `(itemOrRow as Row<T>).data ? ...` branch as misclassifying plain items that happen to have a `data` property. **That was wrong** — the old branch and `getRowData` are logically identical in every case: when `.data` is falsy you get the item; when `.data` is truthy, `getRowData` still requires `.table` and otherwise returns the item unchanged.

`getKey` now delegates to `getRowData` anyway, as a consistency refactor with a truthful comment — but this PR fixes no behavior there, and the test for it is a guard, not a reproduction.

Genuinely still open (deliberately not done here): to mis-key an item you'd need it to carry **both** `data` and `table`, and `getRowData` misfires on that too. Hardening row detection properly needs a brand/`instanceof` check.

## Tests

6 added to the existing Selection module; the 4 bug reproductions confirmed failing first.

- does not hijack Enter or Space from a button inside a cell — **failed** with `defaultPrevented` `true,true`; now `false,false`, the row's `data-selected` stays `false`, and the button's own handler still fires
- does not hijack arrow keys from an input inside a cell — **failed** with `document.activeElement` being an `HTMLTableRowElement` (focus had jumped to row 1)
- uses a roving tabindex so the table has a single tab stop — **failed** with all three rows at `tabindex="0"`; now row 1 only, and the `0` moves to row 2 after ArrowDown
- puts the roving tabindex on the first selected row — **failed** with 3 tabbable rows
- does not add row tab stops when selection is disabled — regression guard
- derives keys correctly for items that have their own `data` property — guard (see above)

On technique: `triggerKeyEvent` does not perform a native button activation, so tests 1–2 prove the fix by capturing `event.defaultPrevented` on a listener above the `<tr>` (the row handler runs first), plus focus and selection state — that is the actual mechanism that would break the control.

Pre-existing coverage verified rather than duplicated: single + multiple selection, select-all (none→all, all→none, indeterminate, with disabled rows), disabled keys, controlled/uncontrolled, row Space/Enter toggling, arrow navigation between rows, and sorting. **No existing assertion was modified** — nothing asserted the buggy behavior.

Full suite: **903 tests, 900 pass, 3 skip, 0 fail** (897 baseline + 6 new), run twice clean. `lint:types` clean.

## Known limitation

Click-focusing a row does **not** move its tab stop — there is no `focusin` sync, deliberately. This component family has a history of focus-during-render flakes (writing tracked state from a handler that also moves focus, where the synchronous `focusout`/`focusin` can land mid-render), and a `focusin` listener is the highest-risk place for exactly that. Arrow navigation moving the tab stop is what the pattern requires; the mouse case is a smaller ergonomic gap left for a follow-up. For the same reason `activeRowKey` is assigned *after* `focus()` in the handler.

`nextRow.dataset['key']` uses bracket access because glint's `TS4111` rejects the dot form — please don't "clean it up".

### DIFF
--- a/packages/frontile/src/components/collections/table.md
+++ b/packages/frontile/src/components/collections/table.md
@@ -1207,11 +1207,19 @@ export default class DemoComponent extends Component {
 
 ### Keyboard Navigation
 
-When selection is enabled, rows can be selected using keyboard:
+When selection is enabled, rows follow the WAI-ARIA grid pattern:
 
+- **Tab**: Moves into the rows. The table is a single tab stop — it uses a
+  roving `tabindex`, so exactly one row is tabbable at a time. That is the first
+  selected row, or the first row when nothing is selected.
+- **Arrow Down** / **Arrow Up**: Move focus between rows, carrying the
+  `tabindex="0"` along with focus.
 - **Space** or **Enter**: Toggle selection (multiple mode) or select row (single mode)
-- Rows are focusable with `tabindex="0"` for keyboard accessibility
-- Disabled rows cannot be selected via keyboard
+- Disabled rows cannot be selected via keyboard, but can still be focused
+- Interactive content inside cells keeps its own keys: a button, link or input in
+  a cell handles Enter, Space and the arrow keys itself, and the row does not
+  toggle its selection. Row keyboard handling only applies to keys pressed on the
+  row itself.
 
 ```gts preview
 import Component from '@glimmer/component';
--- a/packages/frontile/src/components/collections/table/table.gts
+++ b/packages/frontile/src/components/collections/table/table.gts
@@ -298,6 +298,11 @@ class Table<
     return keys ? new Set(keys) : new Set();
   }
 
+  // Roving tabindex (WAI-ARIA grid pattern): the key of the row the user last
+  // moved to with the keyboard. `undefined` until they move — see
+  // `rovingRowKey` for the default tab stop.
+  @tracked activeRowKey?: string;
+
   // Extract actual data from row (handles both row.data and direct row)
   getRowData = (row: Row<T>): T => {
     if (row.data && row.table) {
@@ -308,10 +313,10 @@ class Table<
 
   // Get key from item or row - unified method
   getKey = (itemOrRow: T | Row<T>): string => {
-    // Extract the actual item from row if needed
-    const item = (itemOrRow as Row<T>).data
-      ? this.getRowData(itemOrRow as Row<T>)
-      : (itemOrRow as T);
+    // Extract the actual item from row if needed. `getRowData` already returns
+    // the argument untouched when it is not a table row, so a plain item that
+    // happens to have a `data` property is not mistaken for one.
+    const item = this.getRowData(itemOrRow as Row<T>);
 
     // Use provided getKey function if available
     if (this.args.getKey) {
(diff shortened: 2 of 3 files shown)


josemarluedke:

Both axes independently converged on one real defect. Everything else is a note.

### 🔴 Must fix — the roving tabindex can leave the table with *zero* tab stops

`rovingRowKey` (`table.gts:339-356`) maps `getKey` over **plain items** from `sortedItems`, while `rowTabIndex` maps `getKey` over **`Row` objects**. For an item carrying both `data` and `table`, `getRowData` unwraps it, so `rovingRowKey` yields `'payload-1'` while `rowTabIndex(row)` yields `'1'`. They never match, `rowTabIndex` returns `'-1'` for every row, and **a selectable table becomes entirely unreachable by keyboard** — strictly worse than the 1-per-row bug this PR fixes.

The underlying `getRowData` ambiguity is pre-existing (and this PR's body says so), but Finding 2's new code is what converts it from a mis-keyed select-all into total loss of the tab stop.

This PR's own new test `it derives keys correctly for items that have their own 'data' property` (`table-selection-test.gts:1256+`) renders exactly that shape (`data: 'payload-1', table: 'users'`) and asserts only `data-key` and `data-selected` — it walks straight past the regression.

**Fix:** make both sides key consistently (derive `rovingRowKey` from the same `Row` objects the template iterates, or tighten `getRowData`'s discrimination), and add `assert.dom('tbody tr[tabindex="0"]').exists({ count: 1 })` to that test so it can't regress again.

- **Mixed call conventions** in one file: `this.getKey(row.data)` (lines 406, 411, 416, 471) vs `this.getKey(row)` (355, 826). Works only because both funnel through `getRowData`; a `rowKey(row)`/`itemKey(item)` pair would remove the ambiguity. Related to the above.
- Spec axis confirms Findings 1 & 2 implemented as specified, Finding 3 correctly claims equivalence only, no scope creep, and the `defaultPrevented`-on-an-ancestor test technique is sound (the row handler is a bubble-phase listener on the `<tr>`, so a container listener does observe its `preventDefault`). All four reproductions fail on `main`.

josemarluedke:
Addressed the zero-tab-stop finding (`0c61886a`).

**Test first.** Extended `it derives keys correctly for items that have their own 'data' property` with `assert.dom('[data-test-id="table-row"][tabindex="0"]').exists({ count: 1 })`. It failed on the branch as-is before I touched any source:

```
not ok 1 - it derives keys correctly for items that have their own `data` property
    actual: Element [data-test-id="table-row"][tabindex="0"] does not exist
    expected: Element [data-test-id="table-row"][tabindex="0"] exists once
```

Confirming the diagnosis: `rovingRowKey` keyed off plain `sortedItems` while `rowTabIndex` keyed off `Row` objects, so an item carrying both `data` and `table` got unwrapped by `getRowData` on one side only — no key ever matched and every row went to `-1`.

**Fix.** `rovingRowKey` now maps `getKey` over `headlessRows`, the same collection the template iterates, so both sides take the identical code path. Chose this over tightening `getRowData` because it is a one-line change confined to the code this PR added, whereas `getRowData` also backs selection and sticky-row checks — the ambiguity stays documented and genuinely open, but it can no longer cost the tab stop. Also corrected the `getKey` comment that claimed a plain item with `data` is never mistaken for a row.

Full suite: **903 tests, 900 pass, 3 skip, 0 fail.** `pnpm --filter frontile lint:types` clean.

PR #519, opened 2026-08-28
fix(overlays): reference-count body scroll lock and fix dialog a11y attrs

josemarluedke:
## Problems

**1. Nested overlays unlocked body scroll prematurely.** `setupContent` set `document.body.style.overflow = 'hidden'` on open and `''` on teardown, with no reference counting and no memory of the previous value.

Open Modal A (body locked) → open Modal B from inside it (locked again) → close B → teardown sets `overflow = ''`, but **Modal A is still open and the page behind it scrolls.** Separately, an app with its own inline `body { overflow: ... }` had it wiped rather than restored. The existing tests only ever exercised a single overlay.

**2. `role="dialog"` without `aria-modal`.** Modal and Drawer both render a focus-trapped dialog but never set `aria-modal="true"`, so screen readers keep exposing the page behind them in browse mode.

**3. Dangling `aria-labelledby`.** Both emitted `aria-labelledby={{this.headerId}}` unconditionally, so with no `Header` rendered the attribute pointed at a non-existent id and the dialog had no accessible name.

## Fixes

**Scroll lock** is now reference counted at module scope, remembering the inline value present before the first lock and restoring it only when the last overlay releases. Each overlay records *whether it took a reference* (`didLockBodyScroll`) rather than re-deriving the `renderInPlace`/`blockScroll` condition at teardown — those args can change while the overlay is open, and an asymmetric lock/unlock would leak the count permanently.

**`aria-modal="true"`** unless `@disableFocusTrap` is set, in which case the attribute is omitted rather than emitting a lie. I verified from `ember-focus-trap`'s source rather than inferring it: `createFocusTrap` only builds the trap, and `activate()` — which installs the document-level listeners — is never called when `isActive` is false. The `el.focus()` in `setupContent` is a one-shot focus move that only runs *when the trap is off*; Tab afterwards walks straight out into the page. So with the trap disabled the page behind genuinely is reachable. Not gated on `@renderInPlace`, where focus is still trapped and the promise still holds.

**`aria-labelledby`** is driven by the yielded `Header` registering itself on insert and deregistering on teardown, so it follows a conditionally rendered header in both directions. A DOM probe for the id was rejected: it goes stale when the header appears or disappears later, and it cannot distinguish a Frontile header from a consumer heading that merely carries the same id.

One implementation subtlety, recorded in a comment: the registration counter is deliberately **untracked**, with only the derived `hasHeader` boolean tracked and only ever written from the callback. A single `@tracked` counter using `+=` reads tracked state inside the header modifier's install frame and trips Glimmer's backtracking assertion (34 failures).

## Behavior change, and the guard for it

A consumer who put the yielded `headerId` on their own heading previously got `aria-labelledby` for free and now must also pass `aria-labelledby={{m.headerId}}` themselves. Trading a dangling reference for a *silently* missing name is not a clear win, so this PR also adds a **development-only `warn`** (stripped from production builds) when a dialog resolves to no accessible name at all — no registered header, no `aria-label`, no `aria-labelledby`. Ids: `frontile.modal.missing-accessible-name`, `frontile.drawer.missing-accessible-name`. The message names both remedies and calls out this exact trap.

The check reads the rendered element for consumer-supplied names, since `...attributes` is invisible to args. Ordering is safe write-then-read: element modifiers install bottom-up, so the header's registration precedes the dialog's check — proven by a test that fails loudly if the read lands first, not assumed. Known limitation: a `Header` that only appears in a *later* render pass warns once for the initial unnamed pass and the warning cannot be retracted. Warn-at-open is the right call there — the dialog *was* unnamed when it opened.

## Tests

22 added; the 10 bug reproductions confirmed failing first (`# tests 80 / # pass 70 / # fail 10` on the `overlays` filter).

Scroll lock (4): nested overlays keep the lock until the last closes; a pre-existing inline `overflow: scroll` is restored rather than blanked; `@blockScroll={{false}}` and `@renderInPlace={{true}}` neither lock nor corrupt the count (regression guards for the symmetry requirement).

Modal + Drawer (6 each, plus 6 warning tests): `aria-modal` present when the trap is active and absent when disabled; no `aria-labelledby` without a header; the attribute appears and disappears as a conditional header is added and removed; a consumer `aria-label` is preserved; a consumer `aria-labelledby` via `...attributes` wins over the header id; the yielded-`headerId`-on-your-own-heading case warns.

**No pre-existing test was modified or deleted** — the two original scroll-lock assertions pass unchanged.

**Output cleanliness:** every pre-existing test already renders a `Header`, so none of them warn — verified mechanically, `grep -c "missing-accessible-name"` over the full suite output is 0. The one test that renders a genuinely unnamed dialog asserts the warning through a `registerWarnHandler` helper that restores console delegation in a `finally`.

Full suite: **919 tests, 916 pass, 3 skip, 0 fail** (897 baseline + 22 new). `lint:types` exits 0; `lint:hbs` unchanged at its 62 pre-existing problems, none in the touched files.

## Note

`previousBodyOverflow` is module-global, so a test that leaves an overlay open past teardown could leak it. All 919 tests pass in sequence, and the `if (bodyScrollLockCount === 0) return;` guard makes an unbalanced unlock a no-op rather than a corruption.

### DIFF
--- a/packages/frontile/src/-private/dialog.ts
+++ b/packages/frontile/src/-private/dialog.ts
@@ -1,0 +1,39 @@
+import { warn } from '@ember/debug';
+
+/**
+ * Warns when a dialog ends up with no accessible name at all — no registered
+ * `Header`, and no `aria-label`/`aria-labelledby` of the consumer's own.
+ * Assistive technology announces such a dialog as just "dialog", and nothing
+ * else surfaces the mistake at runtime.
+ *
+ * Consumer-supplied names arrive through `...attributes`, which the component
+ * cannot see in its args, so this inspects the rendered element for them. A
+ * heading merely carrying `headerId` deliberately does *not* count: nothing
+ * points `aria-labelledby` at it, so the dialog really is unnamed, and that is
+ * the case most worth surfacing.
+ *
+ * `warn` is stripped from production builds, so this costs consumers nothing.
+ */
+export function warnIfDialogHasNoAccessibleName(
+  element: Element,
+  hasRegisteredHeader: boolean,
+  component: 'modal' | 'drawer'
+): void {
+  const hasName =
+    hasRegisteredHeader ||
+    !!element.getAttribute('aria-label') ||
+    !!element.getAttribute('aria-labelledby');
+
+  const name = component === 'modal' ? 'Modal' : 'Drawer';
+  const block = component === 'modal' ? 'm' : 'd';
+
+  warn(
+    `<${name}> has no accessible name: assistive technology will announce it ` +
+      `as just "dialog". Render the yielded <${block}.Header>, or pass your ` +
+      `own aria-label or aria-labelledby to <${name}>. Putting the yielded ` +
+      `headerId on a heading of your own is not enough on its own — pass ` +
+      `aria-labelledby={{${block}.headerId}} alongside it.`,
+    hasName,
+    { id: `frontile.${component}.missing-accessible-name` }
+  );
+}
--- a/packages/frontile/src/components/overlays/drawer.gts
+++ b/packages/frontile/src/components/overlays/drawer.gts
@@ -1,4 +1,5 @@
 import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
 import { guidFor } from '@ember/object/internals';
 import { hash } from '@ember/helper';
 import Overlay, { type OverlaySignature } from './overlay';
@@ -94,7 +95,7 @@ export interface DrawerSignature {
         >;
         Header: WithBoundArgs<
           ComponentLike<DrawerHeaderSignature>,
-          'labelledById' | 'classFromParent'
+          'labelledById' | 'classFromParent' | 'registerSelf'
         >;
         Body: WithBoundArgs<
           ComponentLike<DrawerBodySignature>,
@@ -113,6 +114,41 @@ export interface DrawerSignature {
 
 export default class Drawer extends Component<DrawerSignature> {
   headerId = `${guidFor(this)}-header`;
+
+  // `aria-labelledby` is only useful if it points at an element that actually
+  // exists — a dangling reference leaves the dialog with no accessible name at
+  // all. The yielded `Header` reports when it is rendered and when it goes
+  // away, which keeps the reference correct even for a conditional header.
+  // The count itself is deliberately untracked: registration happens while the
+  // header's modifier is installing, and reading a tracked value there before
+  // writing it trips Glimmer's backtracking assertion. Only the derived flag is
+  // tracked, and it is always written, never read, from the callback.
+  renderedHeaders = 0;
+  @tracked hasHeader = false;
+
+  registerHeader = (isRendered: boolean): void => {
+    if (this.isDestroying || this.isDestroyed) {
+      return;
+    }
+
+    this.renderedHeaders += isRendered ? 1 : -1;
+    this.hasHeader = this.renderedHeaders > 0;
+  };
+
+  get labelledById(): string | undefined {
+    return this.hasHeader ? this.headerId : undefined;
+  }
+
+  /**
+   * `aria-modal="true"` promises assistive technology that the rest of the page
+   * is unreachable while this dialog is open. That promise only holds while the
+   * overlay's focus trap is active, so when the consumer turns the trap off we
+   * omit the attribute rather than lie about it — screen readers then keep
+   * exposing the page behind, which is what actually happens.
+   */
+  get ariaModal(): 'true' | undefined {
+    return this.args.disableFocusTrap === true ? undefined : 'true';
+  }
 
   get preventClosing(): boolean {
     return this.args.allowClosing === false;
@@ -171,7 +207,8 @@ export default class Drawer extends Component<DrawerSignature> {
         class={{this.classes.base class=@classes.base}}
         tabindex="0"
         role="dialog"
-        aria-labelledby={{this.headerId}}
+        aria-modal={{this.ariaModal}}
+        aria-labelledby={{this.labelledById}}
         ...attributes
       >
         {{#if this.showCloseButton}}
@@ -192,6 +229,7 @@ export default class Drawer extends Component<DrawerSignature> {
             Header=(component
               DrawerHeader
               labelledById=this.headerId
+              registerSelf=this.registerHeader
               classFromParent=(this.classes.header class=@classes.header)
             )
             Body=(component
--- a/packages/frontile/src/components/overlays/drawer.gts
+++ b/packages/frontile/src/components/overlays/drawer.gts
@@ -2,6 +2,8 @@ import Component from '@glimmer/component';
 import { tracked } from '@glimmer/tracking';
 import { guidFor } from '@ember/object/internals';
 import { hash } from '@ember/helper';
+import { modifier } from 'ember-modifier';
+import { warnIfDialogHasNoAccessibleName } from '../../-private/dialog';
 import Overlay, { type OverlaySignature } from './overlay';
 import DrawerBody, { type DrawerBodySignature } from './drawer/body';
 import DrawerFooter, { type DrawerFooterSignature } from './drawer/footer';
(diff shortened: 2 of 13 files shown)


josemarluedke:

Spec axis: nothing missing, no scope creep outside the accepted list, and the reference counting / `didLockBodyScroll` symmetry / `aria-modal` gating / conditional `aria-labelledby` all verified against tests that fail on `main`. Two things to fix, both in test infrastructure or comments rather than the fix itself.

### 🔴 Must fix — the warning-capture helper leaks across the whole suite

`modal-test.gts:20-45` (and the drawer equivalent): `registerWarnHandler` **cannot be deregistered**. "Restoring" installs a pass-through whose `next` is still the capturing handler, and that handler `return`s without calling `next` for any `frontile.*` id.

So after the first warning test runs, **every Frontile warning for the rest of the suite run is silently swallowed** into an abandoned array, and the handler chain grows per test. No false pass in this PR, but it can mask a genuine warning in an unrelated test — exactly the kind of thing that makes a future regression invisible.

**Fix:** make the capture opt-in per test via a flag the restore actually clears (so the installed handler becomes a true pass-through), and hoist it to `test-app/tests/helpers/` — it is currently 27 lines duplicated verbatim across `modal-test.gts` and `drawer-test.gts`, and that directory already exists.

### 🟡 Should fix — the ordering comment states the wrong reason

`modal.gts:143-150` (identical in `drawer.gts`) says *"element modifiers install bottom-up, so the header's registration has already happened by the time we run"*. Child-before-parent modifier install is an unspecified Glimmer implementation detail, not a contract.

What actually keeps the write safe is that modifier install runs **after the render transaction closes** — the template already consumed `hasHeader` via `this.labelledById` before any modifier installs. The parent/child order only decides whether the *first-render warning* is correct; if it ever inverted, the failure mode is a spurious dev warning on a modal that does have a `Header`, not a crash.

Please reword so the next reader isn't relying on the wrong guarantee. (I asked for this ordering to be verified rather than assumed, and it was — the conclusion is just attributed to the wrong mechanism.)

- **Untested path:** both `aria-labelledby` tests render a `Header`, so "a consumer-supplied `aria-labelledby` suppresses the warn" is never exercised without one.
- `overlay.gts` now owns process-wide mutable module state (`bodyScrollLockCount`, `previousBodyOverflow`) that no test can reset. This PR itself demonstrates the alternative (`-private/dialog.ts`, with `-private/dom.ts` as precedent). Deliberate and guarded, so not blocking.
- The Modal/Drawer duplication (~35 lines each, plus identical `register` modifiers) was weighed and largely suppressed: on `main` those two files are 210 vs 213 lines differing almost only by naming, so near-identical pairs are the established shape here.
- `-private/dialog.ts`: `const block = component === 'modal' ? 'm' : 'd'` — `blockParamName` would be honest.

josemarluedke:
Addressed all three findings.

**1. Warning-capture leak (🔴).** Confirmed before fixing: with a probe registered below the old handler, a `frontile.*` warning raised *after* "restore" was observed as `[]` — swallowed, exactly as described. `registerWarnHandler` is install-only, so the old restore stacked a pass-through *above* the still-capturing handler.

Now one handler pair installs once, and each callback opens a capture *window* that the restore actually closes; outside a window it delegates to `next`. A new test in both suites asserts a post-restore `frontile.*` warning reaches the handler below — it fails against the old helper, passes now. Helper hoisted to `test-app/tests/helpers/frontile-warnings.ts`, dropping 27 verbatim-duplicated lines.

**2. Ordering comment (🟡).** Reworded in `modal.gts`/`drawer.gts`: what is load-bearing is that modifier install runs after the render transaction closes, not bottom-up install order. That order is now called out as unspecified, with its benign failure mode — a spurious dev warning, not a crash.

**3. Untested path (🟡).** Added the consumer-supplied `aria-labelledby` with no `Header` case for Modal and Drawer.

Full suite: **923 tests, 920 pass, 3 skip, 0 fail.** `grep -c missing-accessible-name` over the output is `0`. `lint:types` clean; `lint:js` adds no new warnings.

josemarluedke:
## Follow-up: use the existing `ref` utility instead of the bespoke registration protocol

Raised by @josemarluedke — and it's right. `ref` ([docs](https://next.frontile.dev/docs/components/utilities/ref.md)) already does exactly what the header-registration machinery here hand-rolls: a `setup` modifier that assigns the element to a tracked `current` on install and clears it on teardown.

**What it replaces:**

| current | with `ref` |
|---|---|
| `renderedHeaders` untracked counter (×2 files) | — |
| `@tracked hasHeader` (×2) | `headerRef.current` |
| `registerHeader(isRendered: boolean)` (×2) | — |
| `@registerSelf?: (isRendered: boolean) => void` internal arg (×2 header signatures) | pass `headerRef.setup` down |
| `register = modifier(...)` wrapper inside each header component (×2) | `{{@ref}}` on the root div |

~40 lines across `modal.gts`, `drawer.gts`, `modal/header.gts` and `drawer/header.gts` collapse to roughly four, and a bespoke boolean-callback protocol disappears from the Header arg surface in favour of a documented public utility.

**The pattern is already proven in this repo**, both halves of it: `select.gts` passes `this.triggerRef.setup` to a child component as a bound modifier arg (line 782) and reads `this.triggerRef.current` during render in the same template (line 850).

It also sidesteps the backtracking hazard by construction rather than by discipline: `Ref.setup` only ever *writes* `current`, so there is no read-then-write inside the callback — which is the whole reason the untracked-counter-plus-tracked-flag split exists here. The long comment explaining that split can go with it (the separate note about *when* modifier install runs, relative to the render transaction, is still worth keeping for the warn check).

**One trade-off, stated rather than buried.** The counter is immune to something `listManager.ts` documents explicitly: Glimmer installs a replacement modifier *before* tearing down the one it replaces. `Ref` clears `current` unconditionally on teardown, so a destroy-and-recreate-in-one-pass swap would leave it `undefined` while a header is still present — and the failure mode is a dialog silently losing its accessible name.

For a single `<m.Header>` that is not a realistic pattern (a `{{#if}}` toggle is a plain add or remove, not a swap), so I judge it acceptable. Flagging it because it is a genuine difference in robustness, not a pure win — if you would rather keep the counter for that reason, that is a defensible call and this comment can be closed as considered-and-declined.

josemarluedke:
## Outcome: `ref` was tried and rejected — with evidence

Answering my own suggestion above: **the refactor does not work, and my "not a realistic pattern" judgement was wrong.**

The full refactor was built first (`-42/+38` across `modal.gts`, `drawer.gts` and both `header.gts` — counter, tracked flag, `registerHeader` and both `register` modifiers gone, replaced by `headerRef = ref<HTMLDivElement>()` yielded as `setupHeader`, typed exactly like `SelectTrigger`'s `triggerRef`). It type-checked and built cleanly. Then the trade-off I'd flagged got probed before committing, using the most ordinary swap pattern available — a header in each branch of an `{{#if}}`, i.e. an edit-vs-view dialog:

```hbs
{{#if this.editing}}<m.Header>Editing</m.Header>
{{else}}<m.Header>Viewing</m.Header>{{/if}}
```

- **With `ref`:** `aria-labelledby` is present before the swap and `null` after it, **permanently**, and the dev warning fires — `WARNING: <Modal> has no accessible name`. The dialog silently loses its accessible name, which is the exact harm this PR exists to prevent.
- **With the counter:** both assertions pass. (Verified by stashing the refactor, rebuilding, and re-running the same probe.)

So this is reachable through a plain, idiomatic template with no exotic constructs — not a pathological case.

**Why it's structural rather than fixable in passing.** `Ref.setup` clears `current` unconditionally on teardown, so it cannot distinguish "header removed" from "header replaced". `ref` answers *what is the current element?*; the dialog needs *does any header exist?*. Glimmer installs the modifiers of newly rendered elements **before** tearing down the ones they replace — the property `utils/listManager.ts` documents at lines 152-160 — so on a swap the new `setup` runs, then the old teardown wipes it. A `+1/-1` count is order-independent by construction.

**Decision:** keep the counter. `08f2d5ad` replaces the guesswork with a comment recording why `ref` was rejected and pointing at the `listManager.ts` reasoning, so the next reviewer asking this question gets the answer instead of re-deriving it. Full suite still 923 tests / 920 pass / 3 skip / 0 fail; `lint:types` clean.

Two options deliberately **not** taken, noted in case either is wanted later:
1. A `refCount`-style utility beside `ref` — Modal and Drawer are already two call sites, so if this pattern recurs a third time it earns its own primitive.
2. Making `Ref` swap-safe (clear `current` only if it still `===` the element being torn down). That would fix the general class of bug, but it changes shared public-utility semantics and needs its own tests — a separate PR, not this one.

PR #520, opened 2026-08-28
fix(core): escape attribute selectors in DOM helpers

josemarluedke:
## Problems

**1. Unescaped, unquoted attribute selector.** `-private/dom.ts` built `doc.querySelector(\`[${attribute}=${value}]\`)` with the value interpolated raw. `getElementById` routes through it as `[id=<value>]`, and the value comes from consumer input — `Portal @target="#some-id"` (the `#` is stripped and the rest passed straight in) and `@target="named-target"` (via `data-portal-for`).

- **Breakage:** any HTML-legal id that is not a CSS identifier throws `SyntaxError: not a valid selector` and takes the render down — `#1foo` (ids may legally start with a digit), `#user.email`, `#a:b`, ids containing spaces or `]`.
- **Selector injection:** `@target="#x], [data-portal-target"` escapes the intended predicate and portals content into an arbitrary element.

The existing tests only ever used the plain identifier `#target`, which is why this was never caught.

**2. Same class of bug in `ListManager.setupItem`,** which resolved `aria-labelledby` with `el.querySelector(\`#${labelId}\`)`. Three separate defects: `aria-labelledby` is a **space-separated list of ids**, so a multi-id value silently became a descendant selector; a non-identifier id threw; and it searched only `el`'s descendants although `aria-labelledby` is **document-scoped**, so an external label produced an empty `textValue` and broke type-ahead in Listbox/Select.

## Fixes

- Attribute selector values are **quoted and escaped**. `getElementById` prefers native `Document.getElementById`, falling back to the escaped selector.
- `CSS.escape` is guarded (`typeof CSS !== 'undefined' && typeof CSS.escape === 'function'`) with a minimal fallback, because this module has an SSR path via the `-document` service and the site prerenders. Because the value is now *quoted*, the fallback is genuinely complete — inside a CSS string only the quote, the backslash, and raw newlines are special — so the SSR path is not a weaker path.
- `setupItem` resolves the **first token** of `aria-labelledby` against `el.ownerDocument`, then falls back to the element's own text.
- The node-walking fallback (used when `querySelector` is absent) no longer returns an element whose attribute value *differs* from the one requested — it previously fell through to the "attribute present" branch.

## Deliberate behavior changes

- **An empty value is now matched literally.** The old truthiness check meant `getElementByAttribute(doc, 'id', '')` fell into the `[id]` branch and returned the first element with *any* id — so `Portal @target="#"` silently portaled into an arbitrary element. "Any element carrying the attribute" is still reachable by omitting the argument, which is what the two real callers do.
- **An unresolvable `aria-labelledby` now falls back to the item's own `textContent`** instead of leaving `textValue` empty. Strictly better, but it is a change beyond pure escaping.

## Tests

7 added; 4 of the 14 Portal tests fail against the unfixed `dom.ts` (re-verified by restoring `main`'s version, rebuilding, running, then restoring the fix).

- `renders inside target element by id when the id is not a CSS identifier` — table-driven over 5 shapes (leading digit, dot, colon, space, closing bracket), each asserted to land in its own element
- `renders nothing for an empty id target`
- `does not portal into an unintended element for an injection-shaped target`
- Listbox (3): `textValue` from a document-scoped `aria-labelledby` (external label, and a multi-id value), proven through type-ahead activating the right option; fallback to own text when the label id doesn't resolve; explicit `@textValue` still wins.

**A vacuous test I had to fix, worth knowing about:** my first empty-target test passed *before* the fix. `assert.dom` only searches inside the testing container, but the element the buggy `[id]` lookup resolved to is the first element in the *document* carrying an id, which sits outside that container — so the portal really did render into an arbitrary element and the assertion never saw it. The test now counts `document.querySelectorAll(...)`, with a comment explaining why it is document-scoped.

Full suite: **904 tests, 901 pass, 3 skip, 0 fail** (897 baseline + 7 new). `lint:types` clean. No pre-existing assertion was modified or removed.

## Coverage the maintainer may want to reclaim

This started with 20 direct unit tests for the DOM helpers, which required adding `-private/**/*.js` to the package's `publicEntrypoints`. That was reverted — it would make every `-private` module a published, addressable entrypoint, so the prefix becomes purely nominal and nothing in there can be renamed without a breaking-change discussion. That's an API-surface decision that shouldn't ride along in a bug fix, and there is no existing precedent for it in the suite.

The representative cases were folded into the Portal integration tests instead. What is no longer covered, if you'd rather expose a `-private` or test-support entrypoint:

1. `getElementByAttribute` called directly at all — its `attribute` parameter is never varied.
2. The no-value branch, covered only indirectly through `hasNestedPortals` and the existing `[data-portal]` assertions.
3. Empty-string value on a non-`id` attribute — `portal.gts` guards the named-target call with `if (name)`, so nothing reaches it.
4. **The `Element`-scope path of `getElementById`** (the `querySelector` fallback taken when `doc` has no `getElementById`). Every Portal call site passes a `Document`, so only the native path is exercised. This is the most significant loss — real production code with no test reaching it.
5. The node-walking fallback used by the SSR `-document` service. Untested before this change and untested now — but note this PR *changed* it (see above), and that correction is unverified by any test.
6. The `CSS.escape`-absent fallback. Chrome always provides `CSS.escape`, so the browser suite only ever takes the native path.

### DIFF
--- a/packages/frontile/src/-private/dom.ts
+++ b/packages/frontile/src/-private/dom.ts
@@ -33,10 +33,42 @@ function childNodesOfElement(
   return children;
 }
 
+/**
+ * Escapes a value for use inside a *quoted* attribute selector.
+ *
+ * Values reaching here come from consumers — `Portal @target="#some-id"`, a
+ * portal target name — and an id is only required to be non-empty and free of
+ * whitespace, so plenty of legal ids (`1foo`, `user.email`, `a:b`) are not
+ * valid CSS identifiers. Interpolating them raw either throws a `SyntaxError`
+ * and takes the render down, or, for a crafted value like
+ * `x], [data-portal-target`, closes the predicate and matches something else
+ * entirely.
+ *
+ * `CSS.escape` is the right tool, but it is a browser global and this module
+ * also runs against the `-document` service when the app is prerendered. In a
+ * quoted string only the quote itself, a backslash, and raw newlines actually
+ * need escaping, so the fallback is both small and complete.
+ */
+function escapeAttributeValue(value: string): string {
+  if (typeof CSS !== 'undefined' && typeof CSS.escape === 'function') {
+    return CSS.escape(value);
+  }
+
+  return value.replace(/["\\]/g, '\\$&').replace(/[\n\r\f]/g, ' ');
+}
+
 export function getElementById(
   doc: Document | Element,
   id: string
 ): null | Element {
+  // `Document` (and the `-document` service standing in for it) can answer
+  // this without a selector at all, which sidesteps escaping entirely.
+  // `Element` has no `getElementById`, so those callers keep the selector and
+  // node-walking paths below.
+  if (typeof (doc as Document).getElementById === 'function') {
+    return id ? (doc as Document).getElementById(id) : null;
+  }
+
   return getElementByAttribute(doc, 'id', id);
 }
 
@@ -45,9 +77,17 @@ export function getElementByAttribute(
   attribute: string,
   value: string | undefined = undefined
 ): null | Element {
+  // Only an omitted value means "any element carrying this attribute". An
+  // empty string is a value like any other and is matched literally —
+  // otherwise `@target="#"` would portal into whatever element happens to
+  // carry an `id` first.
+  const hasValue = typeof value !== 'undefined';
+
   if (doc.querySelector) {
-    if (value) {
-      return doc.querySelector(`[${attribute}=${value}]`);
+    if (hasValue) {
+      return doc.querySelector(
+        `[${attribute}="${escapeAttributeValue(value as string)}"]`
+      );
     } else {
       return doc.querySelector(`[${attribute}]`);
     }
@@ -60,9 +100,9 @@ export function getElementByAttribute(
     node = nodes.shift();
 
     if (node && isElement(node)) {
-      if (value && node.getAttribute(attribute) === value) {
+      if (hasValue && node.getAttribute(attribute) === value) {
         return node as Element;
-      } else if (node.getAttribute(attribute)) {
+      } else if (!hasValue && node.getAttribute(attribute)) {
         return node as Element;
       }
     }
--- a/packages/frontile/src/-private/dom.ts
+++ b/packages/frontile/src/-private/dom.ts
@@ -33,6 +33,9 @@ function childNodesOfElement(
   return children;
 }
 
+// eslint-disable-next-line no-control-regex
+const CONTROL_CHARACTERS = /[\u0000-\u001f\u007f]/g;
+
 /**
  * Escapes a value for use inside a *quoted* attribute selector.
  *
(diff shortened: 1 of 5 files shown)


josemarluedke:

Both prohibitions verified clean: `git diff main...fix/dom-selector-escaping -- packages/frontile/rollup.config.mjs` is **empty**, and `test-app/tests/unit/core/dom-test.ts` does not exist on the branch. Coverage was folded into Portal/Listbox integration tests as required. Spec axis found nothing missing and no scope creep outside the accepted list.

Also confirmed non-vacuous, which was the thing I most wanted checked: `portal.gts`' `destinationElement` returns the `getElementById` result directly for a `#`-prefixed target with no fallback, so on `main` the truthiness check really does degrade `#` to `[id]` and portal into the first id-bearing element. Asserting via `document.querySelectorAll` rather than `assert.dom` correctly avoids the container-scoping trap that made the first version of that test pass for the wrong reason.

### 🟡 Should fix — the SSR fallback mangles newlines instead of escaping them, and the comment says "complete"

`-private/dom.ts:57`:

```js
return value.replace(/["\\]/g, '\\$&').replace(/[\n\r\f]/g, ' ');
```

Substituting a space for the character means a value containing a newline silently queries for a **different** value — wrong element or none, with no error. The comment above it claims the fallback is *"both small and complete"*. Correct minimal escapes inside a quoted string are the `\a` / `\d` / `\c` hex forms.

Low severity (ids can't contain whitespace; a named `@target` can), but the comment overstates and this is the prerender-only path, i.e. the one nothing in the browser suite exercises.

### 🟡 Should fix — the two branches escape for different grammars

Same function: `CSS.escape` is an **identifier** escaper, while the call site interpolates into a **quoted string** (`[${attribute}="${...}"]`). Both happen to work — `CSS.escape('1foo')` → `\31 foo`, which a CSS string un-escapes back to `1foo` — so this is not a bug. But the doc comment describes only the *fallback's* regime and then the primary branch uses another.

Either drop the `CSS.escape` branch (the hand-rolled one is genuinely complete for quoted strings, once the newline handling above is fixed) or say in the comment why identifier-escaping is safe inside quotes. Dropping it would also remove the untested-path problem, since Chrome always takes the `CSS.escape` branch and the fallback is never exercised by the suite.

### ⚪ Note (no action requested)
Two exports of the same module now disagree about what an empty value means: `getElementById(doc, '')` returns `null` via the new short-circuit, while `getElementByAttribute(el, 'id', '')` matches `[id=""]` literally per the accepted decision. Both are reachable (Document vs Element receivers). Worth one line of comment recording the intent rather than a change.

josemarluedke:
Pushed `4fd3c9c6`.

**Finding 1** — the fallback substituted a space for newline/CR/FF, which silently queries for a *different* value: wrong element or none, no error. Now emits the CSS hex escapes (`\a `, `\d `, `\c `, plus the rest of C0 and DEL) with their terminating space.

**Finding 2** — took option 1: dropped the `CSS.escape` branch. Every call site interpolates into a quoted string, so the identifier escaper was the wrong grammar for the position even though it round-trips; keeping both meant carrying two escapers and leaving the hand-rolled one — the only correct one here — unexercised by the browser suite. One path now runs everywhere, prerender included, and the doc comment says why.

**Finding 3** — comment on `getElementById` recording that an empty id is *no id* while an empty value in `getElementByAttribute` matches literally, by design.

**Coverage** — `-private/` isn't a rollup public entrypoint and bundles into a hashed chunk, so no unit import without touching `rollup.config.mjs`. Covered through the public path instead: a named `Portal @target` whose name contains a newline, with a space-named decoy target ahead of it in document order. Verified red against the old escaper, green now.

Full suite: 905 tests, 902 pass, 3 skip, 0 fail. `lint:types` clean; `rollup.config.mjs` untouched.

PR #521, opened 2026-08-28
fix(notifications): make timer pause idempotent and removal single-shot

josemarluedke:
## Problems

**1. `Timer.pause()` was not idempotent.** `this.remaining -= Date.now() - this.start` with no guard. The card wires `pause` to `mouseenter` and `resume` to `mouseleave`, so a re-render or a nested enter double-subtracts; enough of it drives `remaining` negative and the next `resume()` fires the timer immediately.

**2. A notification could be removed twice.** `manager.remove()` didn't check `isRemoving`, so a double-click on the close button scheduled two removals and the `later` callback invoked `onRemove` — and therefore the consumer's `@onDismiss` — **twice for one notification**.

**3. `later` writing tracked state after teardown.** The `later` in `notification-card.gts`'s `transitionIn` (setting `hasEntered`) was never cancelled, nor were the two `requestAnimationFrame` callbacks that touch `element`.

**4. The container's `@onDismiss` was bound once and clobbered other containers.** The constructor registered `this.args.onDismiss` a single time, so a later change to the argument was ignored; and the destructor cleared the service's single callback slot unconditionally, so with two containers mounted the second clobbered the first and the first's teardown silently disabled the second's.

**5. Stale `marginTop`.** `setupSpacing` set `marginTop` only for top placements and never cleared it, so switching `@placement` from `top-*` to `bottom-*` left a stale margin.

## Fixes

- `pause()` no-ops when not running and clamps `remaining` at 0. `resume()` no-ops when already running — it previously called `clear()` + `setup()`, resetting `start` and rescheduling the full undecremented `remaining`, so repeated `mouseleave` without `mouseenter` could keep a notification alive indefinitely.
- `remove()` bails when the notification is already being removed, making removal and `onRemove` single-shot.
- `transitionIn` returns a teardown that `cancel()`s the `later` and `cancelAnimationFrame`s both frames.
- The container registers a **stable** `handleDismiss` that delegates to the current `this.args.onDismiss`, so argument changes are picked up without re-registering; the destructor clears the service slot only if it still owns it.
- `setupSpacing` always assigns `marginTop` (empty string for bottom placements) and clears it on teardown.

## Tests

12 added (35 in the notifications filter); 9 confirmed failing first, with the actual wrong values: `remaining` 1697 vs 1899 for the double-pause, `remaining` of **-92** for the negative case, 1799 for the restarted resume, 2 and 3 `onRemove` calls where 1 and 2 were expected, a stale `16px` margin, the old callback being used, and a surviving container getting nothing.

Also added coverage for behavior that had none: the auto-removal `onRemove` path, the `skipTimer` config option, and top-placement reverse ordering (all three passed before — they are guards).

**One honest caveat.** `it can be destroyed while the enter transition is in flight` **passes both before and after the fix.** Ember 6.6 does not raise on tracked writes to a destroyed component, and I could not construct a teardown that races the pending `later` through the test helpers — a timing-based attempt was contaminated by `ember-css-transitions`' own waiter. The test is labelled in-file as a smoke test, and **fix 3 is correct-by-construction rather than test-proven.** Cancelling a timer you own on teardown is unambiguously right, but it is not demonstrated by a failing-then-passing test like the others.

Full suite: **909 tests, 906 pass, 3 skip, 0 fail** (897 baseline + 12 new). `lint:types` clean. No pre-existing test asserted the buggy behavior, so nothing was modified or deleted.

## Scope note on the multi-container case

Staleness is fully fixed, and tearing down a *non-owning* container no longer disables the owner. The reverse — destroying the owning container while another is still mounted — still leaves no callback registered. Fixing that properly needs a multi-subscriber service, i.e. the redesign that was out of scope here. Documented in `packages/frontile/docs/notifications-usage.md` instead, along with `@onDismiss` firing exactly once per notification and the argument being read live.

(`packages/notifications/docs/notifications-usage.md` is a stale wrapper copy that Docfy does not render; left alone.)

### DIFF
--- a/packages/frontile/docs/notifications-usage.md
+++ b/packages/frontile/docs/notifications-usage.md
@@ -315,6 +315,8 @@ export default class PersistentExample extends Component {
 
 Track notification dismissals for analytics, backend updates, or cleanup. When using a global container, set up callbacks in your application template:
 
+`@onDismiss` is called **exactly once per notification**, after it has actually been removed — repeated dismissals of the same notification (e.g. a double-click on the close button) do not call it again. The callback is always read from the current `@onDismiss` argument, so changing it at runtime takes effect immediately.
+
 ```hbs
 {{! app/templates/application.hbs }}
 <NotificationsContainer
@@ -527,8 +529,12 @@ Options for the NotificationsContainer component:
   - `'top-left'` | `'top-center'` | `'top-right'`
   - `'bottom-left'` | `'bottom-center'` | `'bottom-right'`
 - **`spacing`**: Gap between notifications in pixels (default: `16`)
-- **`onDismiss`**: Callback function called when notifications are dismissed
+- **`onDismiss`**: Callback function called once per notification, after it is removed
 - **`class`**: Custom CSS classes for styling
+
+> **Note**: The notifications service holds a single `onDismiss` slot. If more than one
+> `NotificationsContainer` is rendered at the same time, the most recently rendered one
+> owns the callback — another reason to render a single global container.
 
 ## Accessibility
 
--- a/packages/frontile/src/-private/manager.ts
+++ b/packages/frontile/src/-private/manager.ts
@@ -65,7 +65,9 @@ export default class NotificationsManager {
   }
 
   remove(notification?: Notification<Record<string, unknown>>): void {
-    if (!notification) {
+    // Removal is single-shot: a notification already on its way out must not
+    // schedule a second removal, otherwise `onRemove` is called twice for it.
+    if (!notification || notification.isRemoving) {
       return;
     }
 
--- a/packages/frontile/src/-private/timer.ts
+++ b/packages/frontile/src/-private/timer.ts
@@ -27,11 +27,22 @@ export default class Timer {
   }
 
   @action pause(): void {
+    // Pausing an already paused timer would subtract the elapsed time twice,
+    // eventually driving `remaining` to zero or below.
+    if (!this.isRunning) {
+      return;
+    }
+
     this.clear();
-    this.remaining -= Date.now() - this.start;
+    this.remaining = Math.max(0, this.remaining - (Date.now() - this.start));
   }
 
   @action resume(): void {
+    // Already counting down; resuming again would restart the remaining time.
+    if (this.isRunning) {
+      return;
+    }
+
     this.clear();
     this.setup();
   }
--- a/packages/frontile/src/components/notifications/notification-card.gts
+++ b/packages/frontile/src/components/notifications/notification-card.gts
@@ -4,7 +4,7 @@ import { modifier } from 'ember-modifier';
 import { tracked } from '@glimmer/tracking';
 import { service } from '@ember/service';
 import { htmlSafe } from '@ember/template';
-import { later } from '@ember/runloop';
+import { later, cancel } from '@ember/runloop';
 import { on } from '@ember/modifier';
 import { fn, concat } from '@ember/helper';
 import { CloseButton } from '../buttons/close-button';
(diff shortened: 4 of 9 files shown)


josemarluedke:

Spec axis verified all five findings, including the two I singled out: the teardown cancellation covers **all three** callbacks (`outerFrame`, `innerFrame`, and the `later`, with a `typeof innerFrame !== 'undefined'` guard for teardown before the outer frame ran), and `handleDismiss` is genuinely stable with a correct ownership check. Timer tests turn on values the old arithmetic produced (`-90` remaining, `1800` after a restarting resume) — none is fix-insensitive.

### 🔴 Must fix — the smoke test breaches the repo's own TDD standard

`AGENTS.md` → "Bug Fixes → 1. Write a failing test first". The new test in `notification-card-test.gts` admits in its own header that it passes with or without the fix, which means the **largest source hunk in this PR** (the `cancelAnimationFrame`/`cancel(timer)` teardown in `notification-card.gts:57-88`) ships with no test that goes red without it.

I previously accepted "correct-by-construction" here because Ember 6.6 doesn't raise on tracked writes to a destroyed component. That was too quick — the reviewer proposed a workable observable, and #524 in this same series uses it successfully:

**Fix:** assert no pending runloop timer / animation frame remains after `clearRender()` — e.g. `getSettledState().hasPendingTimers === false`, which is exactly how #524 proves its `teardown()` drops queued work. Spying `cancelAnimationFrame` is the fallback. If neither can be made to fail on `main`, say so explicitly in the PR body and drop the test rather than keeping one that implies coverage it doesn't have.

### 🟡 Should fix — the stale duplicate doc is worth deleting separately

Editing `packages/frontile/docs/notifications-usage.md` alone was the **right call** — `site/docfy.config.mjs` sources only `root: '../packages/frontile'`, `pattern: 'docs/**/*.md'`, so `packages/notifications/docs/` is in no source glob, and AGENTS.md's wrapper-package rule points the same way.

But the duplicate is already dead *and* stale on `main`: it carries the pre-v0.18 numbered token `bg-success-50` and a `@package="notifications"` Signature arg. A separate one-line deletion PR would stop the next author diffing the wrong file. Not chargeable to this PR.

- **Encapsulation:** the destructor reaches past the service's setter to compare identity on its field (`this.notifications.onRemoveCallback === this.handleDismiss`). The service offers `setOnRemoveCallback` but no matching read/unregister, so the container compensates by inspecting service state. An `unregisterOnRemoveCallback(cb)` would put the single-slot ownership rule inside the service that owns it. Out of scope; noting that this PR chose documentation over encapsulation.
- Two small items outside the finding list, both benign: the new `setupSpacing` destructor clearing `marginTop` (a no-op, since the element is being destroyed), and `Math.max(0, …)` in `timer.ts` (defence-in-depth; the `isRunning` guard is the actual fix).
- Finding 2's test drives `service.remove()` twice directly rather than double-clicking the close button. Same guard, one layer below the reported reproduction — not vacuous.

josemarluedke:
**Path taken: replaced the smoke test with a discriminating one.**

`getSettledState().hasPendingTimers` turned out unusable as the observable: the card also renders `{{cssTransition}}`, and ember-css-transitions schedules its own runloop timer for the (transition-delay + transition-duration) window, so a timer is pending after teardown either way. The new test instead renders without awaiting settled (transition still in flight), tears the card down, and asserts on the two cancellations directly — a `cancelAnimationFrame` spy, and backburner's `cancel` counter from `getSettledState().debugInfo`.

Reverting only the teardown hunk in `notification-card.gts`, both assertions go red:

```
not ok 8 - ...: it cancels the in-flight enter transition when destroyed
  message: teardown should cancel the pending enter timer; 0 runloop timer(s) were cancelled
  message: teardown should cancel the pending animation frame(s) of the enter transition; cancelAnimationFrame was called 0 time(s)
```

Fix restored. Full suite: **909 tests, 906 pass, 3 skip, 0 fail**. `frontile lint:types` clean. Commit `d7c883a`.

PR #522, opened 2026-08-28
fix(popover): correct destroy guard, didClose timing and modifier keys

josemarluedke:
## Problems

**1. The destroy guard was inverted.**
```js
if (!this.isDestroyed || !this.isDestroying) { this.isClosing = false; }
```
`||` should be `&&`. While destroying (`isDestroyed === false`, `isDestroying === true`) the condition is `true`, so it writes tracked state on a destroying component — exactly what the guard exists to prevent.

**2. `@didClose` fired before the close transition.** The arg is documented as "Callback when closing has finished, including any exit transition", but `close()` invoked it synchronously, so a consumer that unmounts content in `didClose` cut the exit animation. The `Content`-level `@didClose` already waited for the Overlay's teardown — the two paths disagreed and the Popover-level one was wrong.

**3. Modifier-key letter presses opened the popover.** The type-ahead branch had no modifier check, so focusing a `Select`/`Dropdown` trigger and pressing Cmd+R, Cmd+L, Cmd+C or Ctrl+F opened the dropdown on top of the page while the browser did its own thing.

**4. Dead code.** `updateAriaExtanded` (sic) was referenced nowhere, and `triggerEl` — verified against `main` across `packages`, `test-app`, `site` and `docs` — had **no reader outside that dead modifier**. Both removed.

## Fixes

`close()` no longer calls `args.didClose()`. It records `isDidClosePending` (only when the popover was actually open) and debounces `resetIsClosing`; the `didClose` method already bound to `Content` as `internalDidClose` — which the Overlay calls from its teardown — clears the flag and fires the consumer callback exactly once. Guard corrected to `&&`. Modifier keys excluded from the letter branch, Shift untouched.

## How the two hard tests were made to fail honestly

**`didClose` timing.** `Overlay.isAnimationEnabled` is `false` under `isTesting()`, so `transitionDuration` is forced to 0 and a non-zero transition is unobtainable in tests. What is still real is that the Overlay schedules `didClose` from its `setupContent` **teardown**. So the test dispatches the closing click with a raw `element.click()` (no `await`), asserts the callback has not run at that point — pre-fix it had, synchronously — then after `settled()` asserts it ran exactly once *and that the content element was already detached from the document when it ran*. That is ordering, not a zero-duration coincidence.

**The destroy guard.** Two dead ends worth recording. Closing and unmounting in one turn and waiting out the 90 ms debounce does **not** reproduce it: backburner has no destroy check, the callback fires, but by then the component is *fully* destroyed, both flags are `true`, and the inverted `||` accidentally blocks the write. And `destroy()` from `@ember/destroyable` doesn't flip Glimmer's flags — `@glimmer/component@2.1.1` keeps them in its own WeakMaps.

The real window comes from `EmberGlimmerComponentManager#destroyComponent`, which sets `DESTROYING` and then only *schedules* `willDestroy` and destruction. So the rest of that runloop runs with `isDestroying === true, isDestroyed === false`. The test patches `willDestroy` to call the real `didClose()` from inside that window and counts writes through a patched `isClosing` descriptor, asserting the captured flags really were `{destroying: true, destroyed: false}` so it cannot silently stop testing the window. Ember 6.6 raises nothing for such a write, which is why this is white-box rather than assertion-based.

## Decisions

**`aria-expanded` — left as the incidental modifier re-run, now documented.** Updating it imperatively from `open()`/`close()` would be cheaper (it currently rebuilds a `ResizeObserver` on every toggle), but it would miss controlled mode: a consumer flipping `@isOpen` from outside never goes through those methods, so `aria-expanded` would go stale. That's a real accessibility regression traded for one avoided observer rebuild. A comment names the tracked read as the deliberate mechanism, and a new test flips `@isOpen` externally in both directions so any future switch to an imperative update fails loudly.

**The 90 ms `isClosing` window — deliberate, left alone.** It absorbs a single event cascade (the Overlay's outside-click handler closing while the trigger's click handler is about to re-open), not the animation. Tying it to the 200 ms transition would leave the trigger feeling dead for a fifth of a second after every close — worse than a re-click inside 90 ms being dropped. Documented in the `resetIsClosing` JSDoc. The `@didClose` fix decouples the two properly: the consumer callback no longer rides on this timer.

## Tests

10 added (8 → 21 in the popover filter); 4 confirmed failing first. Guard tests were added for behavior that had **no** coverage at all: Escape closes, ArrowDown/ArrowUp open, Tab closes without focus restore, a bare letter and Shift+letter still open, `aria-expanded` tracks state including external `@isOpen` changes.

Full suite: **907 tests, 904 pass, 3 skip, 0 fail** (897 baseline + 10 new). Filtered: popover 21/21, dropdown 15/15, select 149 pass/1 skip, autocomplete 28/28. `lint:types` clean. No pre-existing test was modified — all 8 original popover tests untouched.

- **The change consumers will feel:** `autocomplete.gts` and `select.gts` both hook `@didClose` to reset internal state (`_inputValue`, `activeDescendant`, search results, filter value). That reset now happens after the Overlay teardown rather than synchronously inside `close()`. All 178 select/autocomplete tests pass, but this is the highest-risk part of the change.
- **A popover whose `Content` never mounts never receives `@didClose`.** Previously it fired unconditionally from `close()`. No such usage exists in-tree (a Popover without `p.Content` renders nothing) and it's documented in the JSDoc, but it is a real narrowing.
- **Re-open inside the transition:** re-opening at ~120 ms still lets the previous cycle's teardown flush `@didClose` while the popover is open again. Pre-existing Overlay behavior — the `Content`-level `@didClose` already worked this way — and left alone, since the close genuinely did complete.
- The white-box test patches `Popover.prototype`, restoring in `finally`. It will break if `didClose` is renamed or `isClosing` stops being a prototype accessor; an `assert.ok` on the descriptor makes that a clear failure rather than a silent pass.

## Separate bug found and deliberately not fixed

`preventFocusRestore` is set to `true` when Tab closes the popover and is **never reset**, so every subsequent close of that same popover also skips focus restoration. Out of scope here; worth its own fix.

### DIFF
--- a/packages/frontile/src/components/collections/dropdown.md
+++ b/packages/frontile/src/components/collections/dropdown.md
@@ -641,13 +641,13 @@ menu is a list of commands rather than a set of choices.
 
 Keys are handled in two places, which is worth knowing when debugging one that doesn't fire:
 
-| Focus is on the trigger | Behavior                                        |
-| ----------------------- | ----------------------------------------------- |
-| `Enter` / `Space`       | Opens the menu, on key release                  |
-| `ArrowDown` / `ArrowUp` | Opens the menu                                  |
-| any letter              | Opens the menu                                  |
-| `Escape`                | Closes                                          |
-| `Tab`                   | Closes and moves on, without pulling focus back |
+| Focus is on the trigger | Behavior                                          |
+| ----------------------- | ------------------------------------------------- |
+| `Enter` / `Space`       | Opens the menu, on key release                    |
+| `ArrowDown` / `ArrowUp` | Opens the menu                                    |
+| any letter              | Opens the menu (not with `Cmd`/`Ctrl`/`Alt` held) |
+| `Escape`                | Closes                                            |
+| `Tab`                   | Closes and moves on, without pulling focus back   |
 
 Once open, focus moves into the menu and the list takes over:
 
--- a/packages/frontile/src/components/overlays/popover.gts
+++ b/packages/frontile/src/components/overlays/popover.gts
@@ -113,8 +113,6 @@ interface PopoverSignature {
 }
 
 class Popover extends Component<PopoverSignature> {
-  triggerEl?: HTMLElement;
-
   /**
    * The element `measureWidth` is watching, when a consumer opts into measuring
    * something other than the trigger. Its presence is what gives `measureWidth`
@@ -123,6 +121,14 @@ class Popover extends Component<PopoverSignature> {
   widthEl?: HTMLElement;
 
   menuId = guidFor(this);
+
+  /**
+   * Whether the consumer is still owed a `@didClose`. Set by `close()` when a
+   * popover that was actually open was closed, cleared by whoever fires the
+   * callback, so it can only ever fire once per close.
+   */
+  isDidClosePending = false;
+
   @tracked _isOpen = false;
   @tracked isClosing = false;
   @tracked preventFocusRestore = false;
@@ -164,6 +170,12 @@ class Popover extends Component<PopoverSignature> {
   };
 
   close = () => {
+    // Only a popover that was actually open has a close to finish, and so a
+    // `@didClose` to owe. `close()` is reachable with nothing open -- a stray
+    // `mouseleave`, a consumer calling the yielded `close` -- and firing the
+    // callback for those would be a lie.
+    const wasOpen = this.isOpen;
+
     this.isClosing = true;
     if (typeof this.args.onOpenChange === 'function') {
       this.args.onOpenChange(false);
@@ -171,16 +183,15 @@ class Popover extends Component<PopoverSignature> {
       this._isOpen = false;
     }
 
-    if (typeof this.args.didClose === 'function') {
-      this.args.didClose();
+    if (wasOpen) {
+      this.isDidClosePending = true;
     }
 
-    debounce(this, this.didClose, 90);
+    debounce(this, this.resetIsClosing, 90);
   };
 
   trigger = modifier(
     (el: HTMLElement, [eventType]: [eventType?: 'click' | 'hover']) => {
-      this.triggerEl = el as HTMLLIElement;
       // The trigger is only the width reference when nothing more specific was
       // nominated. `widthEl` is checked when the measurement happens rather than
       // when the modifier installs, so the two modifiers can install in either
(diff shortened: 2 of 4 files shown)


josemarluedke:

Standards axis: **clean** — no findings, no smells. Spec axis found one real bug.

### 🔴 Must fix — controlled mode can arm a spurious `@didClose`

`popover.gts:172-186`: `close()` sets `isDidClosePending = true` whenever `wasOpen`, but in controlled mode it only calls `onOpenChange(false)`. If the consumer **declines** to close, the content stays mounted and the flag stays armed. It is cleared only in `didClose`, never on re-open — so the next time the content unmounts for *any* reason (route teardown of a surrounding `{{#if}}`, a different close path), `@didClose` fires for a close the consumer never performed.

That is the same class of "would be a lie" the surrounding comment is guarding against.

**Fix:** reset `isDidClosePending = false` in `open()`, plus a test that a controlled popover whose consumer ignores `onOpenChange(false)` does not later emit `@didClose`.

- Prohibition verified: the `trigger` modifier was **not** rewritten — its only changes are the two deleted `triggerEl` assignments, the modifier-key guard, and a comment.
- Dead-code removal verified independently: after deleting `updateAriaExtanded`, the only remaining `triggerEl` references anywhere in `packages`/`test-app`/`site`/`docs` are in the committed build artifact `packages/frontile/declarations/.../popover.d.ts`. Nothing was ever yielding it.
- The tracked-write test is a genuine repro, not a guard: it monkeypatches the tracked setter, asserts `writesWhileDestroying === 0`, **and** asserts the callback really ran with `{destroying: true, destroyed: false}` — so it cannot silently stop testing the window.
- Splitting `didClose` into `resetIsClosing` + `didClose` is an improvement on `main`, where one method did both.

josemarluedke:
Addressed the spurious-`@didClose` finding.

**Test first.** Added `a declined close in controlled mode does not leave @didClose armed` to `popover-test.gts`: a controlled popover whose `onOpenChange` refuses to close, then re-opened, then torn down from outside via `@isOpen`. Pre-fix it failed with `actual: 1, expected: 0` — "@didClose did not fire for the close the consumer declined".

**Fix.** `open()` now clears `isDidClosePending`; opening means the close it was armed for never completed. Updated the field's doc comment, which claimed only `didClose` clears it.

**One correction to the finding:** teardown of a surrounding `{{#if}}` does *not* leak the callback — `Overlay`'s exit `later` is guarded by `!this.isDestroyed`, so it never fires. I verified that separately. The observable leak is the externally-driven unmount, which is what the test covers.

Full suite: 908 tests, 905 pass, 3 skip, 0 fail. `pnpm --filter frontile lint:types` clean.

PR #523, opened 2026-08-28
fix(forms): preserve numeric segments in validation error paths

josemarluedke:
## Problems

**1. Validation issue paths dropped numeric index `0` and never matched array fields.** Two places did the same thing — `filterFieldIssues` in `utils/standard-validator.ts` and `validatorToFormErrors` in `components/forms/form.gts` — mapping the Standard Schema `issue.path` segments and then calling `.filter(Boolean)`:

- `.filter(Boolean)` **removes the numeric index `0`** (and any empty-string key), so `['items', 0, 'name']` became `items.name`.
- `filterFieldIssues` then compared `key === fieldParts[i]`, i.e. `number === string`, which is never true for an index — so per-field validation for `items.0.name` could never match its issue.

**2. The error feedback live region was announced unreliably.** `aria-live` sat on the feedback element, but `FormControl` only renders that element *once the field becomes invalid*. A live region inserted at the same moment as its content is generally not announced — the region has to already exist for assistive technology to be observing it.

## Fixes

Each resolved path segment is coerced with `String(...)` and only `null`/`undefined` are dropped, in both places, with a comment noting the two must stay in sync. That fixes both defects at once: index `0` survives, and the comparison is string-vs-string so any index matches.

`FormControl` now owns a persistent, visually hidden `aria-live="assertive"` region that is always in the DOM and merely filled, and passes `@announce={{false}}` to both the automatic and the yielded `Feedback` so nothing is announced twice. `FormFeedback` gains an `@announce` arg (default `true`, so the standalone public component is unchanged).

**Why not simply always-render the existing feedback element**, which is the obvious reading of "keep the region present and empty":
1. **Layout.** The theme's `formFeedback` base includes `pt-1`, so an always-present empty div adds 4px of padding to every form control. Stripping classes when empty then breaks the block-content case.
2. **Blast radius.** 73 assertions across 11 test files use the presence/absence of `[data-component="form-feedback"]` as *the* idiom for "is there an error". Rewriting them to assert emptiness would have rewritten the meaning of the suite's error assertions and buried the actual fix.

The region is deliberately **not** part of `describedBy` — it is an announcement channel, not a description, and the visible feedback is what should describe the control. Always-rendering it removes a dangling-reference risk rather than adding one.

## Docs: one escape hatch worth knowing

Because the region only ever carries the `@errors` text, a yielded `c.Feedback` with custom *block* content would no longer be announced. The hatch is to pass `@announce={{true}}` at the invocation site, which beats the curried `announce=false`. That precedence claim was **verified with a throwaway probe test before being documented**, not taken on trust — and the probe was then promoted into two permanent tests, since a documented escape hatch with nothing guarding it is how docs rot.

## Tests

11 added; 7 confirmed failing first (re-verified afterwards by checking out just the two path fixes, rebuilding, and re-running: `# tests 3 / # pass 0 / # fail 3`).

- `filterFieldIssues matches array index path segments` — `['items',0,'name']` matches `items.0.name`; `[{key:'items'},{key:2},{key:'name'}]` matches `items.2.name`; indexed issues do not leak onto `items.name`
- `validateField works with array element field names` — through a real valibot schema, indices 0 and 2 report errors and index 1 reports none
- `filterFieldIssues ignores null and undefined path segments` — guards the `.filter(Boolean)` → explicit-nullish swap
- End-to-end `<Form>`: array element errors surface on the indexed field name and not on `items.name`, both `<form.Field>`s render their own message, and index 1's clears on change
- The live region exists and is empty *while valid*, then fills; the visible feedback has no `aria-live` (no double announcement); it empties again on return to valid; it joins multiple messages; it is still rendered with `@preventErrorFeedback`
- The `@announce` override and the curried default
- `it collects multiple issues for the same field into a list of messages` — a coverage gap found in review; passed before, documents existing correct behavior

Unit tests use **valibot**, already used throughout that file. Integration tests use the `@validate` custom-validator path. **No dependency added.**

Full suite: **908 tests, 905 pass, 3 skip, 0 fail** (897 baseline + 11 new). `lint:types` clean. **No pre-existing test was modified or deleted** — only appends and two import lines.

## A correction to how this bug was originally reported

I described the failure through a `z.object({ items: z.array(...) })` schema. **That scenario is unreachable, independent of this fix.** Integration tests written that way failed *even with the fix applied*: `Form` rebuilds nested data by splitting dotted names, and a numeric segment becomes an ordinary object key, so `items.0.name` / `items.1.name` arrive as `{ items: { 0: …, 1: … } }` — an object, never an array. A schema declaring `items` as an array fails on `items` itself with a type error and no per-index issue is ever produced.

The path fix is still correct and still matters — for the `@validate` path, and for any consumer emitting issues with index segments. But array *schemas* don't work end-to-end for a separate reason, and `form.md` now says so unambiguously rather than hedging. Making them work means teaching `unflattenData` to reconstruct arrays, which changes `@onChange`/`@onSubmit` payload shapes — a larger, potentially breaking change that deserves its own issue.

## Note

The error text is now duplicated in the DOM (visible + `sr-only`). That is the accepted cost of this pattern; no test regressed on it. `Checkbox`/`Radio`/`Switch` use `@preventErrorFeedback` plus a yielded `c.Feedback` and are covered by the new region with no change to those files.

### DIFF
--- a/packages/frontile/src/components/forms/field.md
+++ b/packages/frontile/src/components/forms/field.md
@@ -10,6 +10,12 @@ A component wrapper that provides conveniences for form fields when using built-
 
 **Important:** The Field component is designed specifically for use with the Form component's built-in validation system (`@schema` or `@validate` props). It should only be used when implementing form validation.
 
+`@name` is looked up verbatim in the form's `errors` object and read out of the form data
+with Ember's `get`, so dotted paths work — including numeric segments for array elements.
+`@name='items.0.name'` picks up an issue whose path is `['items', 0, 'name']`; the index is
+part of the key, so `items.0.name` and `items.1.name` never collide, and neither collapses
+onto `items.name`.
+
 ## Import
 
 ```js
--- a/packages/frontile/src/components/forms/form-control.gts
+++ b/packages/frontile/src/components/forms/form-control.gts
@@ -4,6 +4,7 @@ import { hash } from '@ember/helper';
 import Feedback, { type FormFeedbackSignature } from './form-feedback';
 import Description, { type FormDescriptionSignature } from './form-description';
 import Label, { type LabelSignature } from './label';
+import VisuallyHidden from '../utilities/visually-hidden';
 import type { ComponentLike, WithBoundArgs } from '@glint/template';
 
 interface FormControlSharedArgs {
@@ -155,6 +156,20 @@ class FormControl extends Component<FormControlSignature> {
     }
   }
 
+  /**
+   * The error messages as a single string, used by the persistent live region.
+   * Matches how `FormFeedback` joins an array of messages.
+   */
+  get errorMessageText(): string {
+    const { errors } = this.args;
+
+    if (!errors) {
+      return '';
+    }
+
+    return typeof errors === 'string' ? errors : errors.join('; ');
+  }
+
   get showErrorFeedback(): boolean {
     if (!(this.args.preventErrorFeedback === true) && this.isInvalid) {
       return true;
@@ -191,6 +206,7 @@ class FormControl extends Component<FormControlSignature> {
             size=@size
             messages=@errors
             intent="danger"
+            announce=false
           )
         )
         to="default"
@@ -202,8 +218,22 @@ class FormControl extends Component<FormControlSignature> {
           @size={{@size}}
           @messages={{@errors}}
           @intent="danger"
+          @announce={{false}}
         />
       {{/if}}
+
+      {{!
+        The visible feedback is only rendered once the field is invalid, and a
+        live region that appears at the same moment as its content is not
+        reliably announced. This region is always in the DOM, visually hidden
+        and empty, so assistive technology is already observing it when the
+        messages arrive. It is intentionally not part of describedBy; the
+        visible feedback is what describes the control.
+      }}
+      <VisuallyHidden
+        aria-live="assertive"
+        data-component="form-feedback-live-region"
+      >{{this.errorMessageText}}</VisuallyHidden>
     </div>
   </template>
 }
--- a/packages/frontile/src/components/forms/form-control.gts
+++ b/packages/frontile/src/components/forms/form-control.gts
@@ -1,7 +1,10 @@
 import Component from '@glimmer/component';
 import { guidFor } from '@ember/object/internals';
 import { hash } from '@ember/helper';
-import Feedback, { type FormFeedbackSignature } from './form-feedback';
+import Feedback, {
+  feedbackMessageText,
+  type FormFeedbackSignature
+} from './form-feedback';
 import Description, { type FormDescriptionSignature } from './form-description';
 import Label, { type LabelSignature } from './label';
 import VisuallyHidden from '../utilities/visually-hidden';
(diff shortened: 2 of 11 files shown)


josemarluedke:

Spec axis: both halves implemented, no new dependency (`package.json` untouched; `valibot` and `ember-resources` were already `test-app` deps), and every accepted decision checks out. The `@announce` precedence claim is guarded by two tests rather than left to the docs — good.

### 🟡 Should fix — the path normalisation is duplicated, with a comment asking a human to keep it in sync

Byte-identical five-line blocks in two files:

- `packages/frontile/src/components/forms/form.gts:535` (`validatorToFormErrors`)
- `packages/frontile/src/utils/standard-validator.ts:233` (`filterFieldIssues`)

…and `form.gts` says the quiet part: *"This must stay in sync with `StandardValidator.filterFieldIssues`."* A comment asking a future reader to keep two copies aligned **is** the smell, and the dotted-name format is precisely the thing that must not diverge — `@name` lookup depends on it, and the original bug was these two disagreeing about index `0`.

**Fix:** extract one `issuePathKeys(path)` / `issuePathToFieldName(issue)` into `standard-validator.ts` and call it from both.

### 🟡 Should fix — the `'; '` join is duplicated too

`form-control.gts:159` re-implements `FormFeedback`'s join with the comment *"Matches how `FormFeedback` joins an array of messages"*, while `form-feedback.gts` keeps its own `messageText`. Same fix: one shared helper, or have the live region reuse `FormFeedback` with `@announce` inverted. A separator living in two places will drift.

### 🟡 Should fix — `@announce` is documented in the wrong place

The new **public** arg on `FormFeedback` is explained in `form-control.md`, but `FormFeedback` has no co-located `.md` and the `## Feedback Messages` section of `form.md` — which documents that component's other args, and which this PR already edits — never mentions it. A consumer rendering `FormFeedback` standalone will look in `form.md`. AGENTS.md → "Documentation": *"Document new props/args and add usage examples here."* A cross-reference is the minimum.

- **The `Field`-side lookup is asserted rather than demonstrated.** The new integration tests read the value through a local `readPath` helper on the validator's argument, not through `Field`'s own `get()` lookup at an indexed path; the verification lands only as prose in `field.md`. Given this codebase's history of helper-shaped vacuous assertions, worth a real test eventually.
- **The live region can still be inserted-with-content on first paint.** A `FormControl` that renders already-invalid inserts region and text in the same render, so the original failure mode survives for that case. A doc caveat is probably enough.
- Not scope creep, contrary to a first-pass reading: the array-schema `form.md` note and the error-text duplication are both in the accepted list.

josemarluedke:
Addressed all three review findings in a83fcf0f.

**Finding 1 (path duplication):** extracted `issuePathKeys(path)` and `issuePathToFieldName(issue)` into `packages/frontile/src/utils/standard-validator.ts` (already a public entrypoint, no rollup change). Both `StandardValidator.filterFieldIssues` and `form.gts`'s `validatorToFormErrors` now call them, so the dotted `@name` format has one definition — the "keep in sync" comment is gone because the coupling is structural now.

**Finding 2 (`'; '` join):** extracted `feedbackMessageText` from `form-feedback.gts` and used it for `FormControl`'s persistent live region. One separator, one place.

**Finding 3 (`@announce` docs):** added `@announce` prose to `form.md`'s `## Feedback Messages`, cross-referencing the fuller explanation in `form-control.md`. No new code fence.

Full suite: **908 tests, 905 pass, 3 skip, 0 fail.** `lint:types` clean. No test assertion was altered.

PR #524, opened 2026-08-28
fix(collections): coalesce ListManager batch work and refresh Listbox callbacks

josemarluedke:
## Problems

**1. Quadratic work and uncancelled timers on every list render.** `ListManager.register` ran once per item and each time:
- called `onListItemsChange(this.#orderedItems, 'add')`, where `#orderedItems` is an uncached getter that filters and `sort()`s every registered item via `compareDocumentPosition`;
- scheduled a `later(…, 1)` that activated the first/selected option, sorting again.

None were cancelled, and `unregister` scheduled another. A `Select` with 500 options performed ~500 sorts of 500 DOM nodes during render, then queued ~500 timers each sorting again — visible jank, and every timer had to drain before `settled()` resolved in consumers' test suites.

**2. Callbacks were captured once and went stale.** `Listbox` constructed `new ListManager({...})` reading `onSelectionChange`/`onAction`/`onActiveItemChange` in a field initializer, and the `setup` modifier never passed them, so `updateArgs` could not replace them. `<Listbox @onAction={{fn this.pick group.id}}>` inside an `{{#each}}` kept calling the **original** closure with the **original** `group.id` forever. Frontile's own callers pass stable arrows, which is why nothing internal caught it.

**3. Multiple tab stops inside the listbox.** `item.gts` returned `tabindex="0"` for every item that was active **or selected**, so a multi-select with 8 selections was 8 tab stops.

Plus small dead/misspelled things in the same files: the `['Enter',' ','Space']` branch (`'Space'` is an `event.code` value and can never match `event.key`), `@edefaultValue 'frist'`, and the private `#clearSeaerch`.

## Fixes

Registration work is **coalesced into one cancellable flush per batch**: one `onListItemsChange` call and one deferred activation per batch instead of N of each. The sort now runs once per batch (O(n log n)) rather than once per registration (O(n² log n)), and it runs after the render against the settled DOM, where the answer is also more accurate.

Callbacks are threaded through the `setup` modifier so `updateArgs` replaces them. Tab stop is a single tracked decision on the manager, which `item.gts` consults.

### `#orderedItems` was deliberately *not* memoized

The suggestion to memoize per render pass was investigated and rejected. The file's own comments explain why: Glimmer **moves** the element of a persisting item with no registration to invalidate a cache, and does so *inside the same synchronous render pass* in which new items register. Any cache keyed on registration or on a render pass can be read after a move that didn't invalidate it — precisely the stale-order bug those comments warn about, and there is no cheap key that tracks DOM moves. Coalescing removed the *need* to recompute instead.

### Teardown

`ListManager` is a plain class, so the hook is the `setup` modifier's destructor calling a new `teardown()`. Two complications handled: a function modifier's destructor also runs *before every re-invocation*, so `teardown()` only cancels the timer and leaves the pending flags, with the `setup` body re-queueing a still-pending flush; and an `#isTornDown` flag covers the reverse destructor order, so a late `unregister` records what it did but queues nothing.

### Roving tabindex vs. the render-timing hazard

This file has a documented history of backtracking-rerender assertions (writing `tabindex` onto the focused item blurs it and dispatches `focusout` synchronously inside the frame that just consumed `isActive`). The change was structured to avoid the hazard rather than tested into safety:

- The decision is **one tracked field on the manager**, not a per-item getter scanning siblings — a scanning getter would consume every item's `isActive` tag, so any mid-render write would dirty tags the frame had just consumed.
- The refresh reads only untracked mirrors (`isSelectedUntracked`/`isDisabledUntracked` added alongside the existing `isActiveUntracked`), consuming nothing.
- The field is written through the same mirror guard as `ListItem`'s fields, so a no-op write never dirties a tag.
- `unregister` deliberately does **not** clear the tab stop; that write would land while Glimmer tears down elements whose surviving siblings have already read it for `tabindex`. The flush corrects it a tick later, during which no option is tabbable — chosen over an assertion, and commented in place.

Both render-safety scenarios pass unchanged; that file needed no modification.

## Tests

12 added; 11 confirmed failing first.

- a replaced `@onAction`/`@onSelectionChange`/`@onActiveItemChange` is the one invoked (Bug 2)
- several selected options still make a single tab stop; the tab stop follows the active option; the first option owns it when nothing is active or selected; a disabled option is skipped by the fallback (Bug 3)
- a batch of registrations reports the list once, in DOM order (was 8 calls); a batch of unregistrations reports once (was 2); a batch that replaces the list reports once (was 5) (Bug 1's observable)
- `teardown` drops the work a batch had queued — asserts `getSettledState().hasPendingTimers` true after a batch, false after `teardown()`

**Honest caveat:** `the deferred activation runs once for the whole batch` **passed before the fix too.** Activation count is not observable — the redundant timers all activated the same item and `activateItem` no-ops when already active. It is kept as a regression guard, not offered as proof.

Full suite: **909 tests, 906 pass, 3 skip, 0 fail** (897 baseline + 12 new). Filters: listbox 28/28, ListManager 14/14, select 153 (1 pre-existing skip), autocomplete 28/28, dropdown 15/15. `lint:types` clean; `lint:js` and `lint:hbs` byte-identical to a stashed baseline. No pre-existing test modified or deleted.

- **`onListItemsChange` timing changed** from synchronous-during-`register` to one deferred call per batch, and a batch that both adds and removes reports the *last* action once rather than one call per mutation. No caller reads the action, and all consumer tests pass, but it is an observable API change. Worth an eye on whether a `Select` with a preselected value can now show its placeholder for an extra tick before the label resolves.
- **`'onAction' in args`** relies on Glimmer's captured named-args dict containing keys written literally in the template even when the value is `undefined`. It works, but it is the load-bearing assumption of the Bug 2 fix.
- `setup` now forwards its whole named-args object to `updateArgs` instead of a hand-picked subset.
- **Two behavior unifications beyond the stated bugs:** the removal path now uses `setFirstOptionActive()` (skips disabled) where it used `activateItem(#firstVisibleItem)` (did not), and the `#items.length > 1` guard was dropped, so a one-item list in `autoActivateMode="selected"` is now auto-activated where it previously was not.

## Not claimed

**"Exactly one tabbable element in the composite" is *not* achieved.** `<ul tabindex="0">` was left in place: `Dropdown`/`Select`/`Autocomplete` render the listbox inside a focus trap with no `fallbackFocus`, and an empty menu with no tabbable node makes `focus-trap` throw. Tab therefore sees two stops (container + one option), down from 1 + N. The roving invariant is now in place, making the container a one-line follow-up once focus-trap fallback focus is sorted.

### DIFF
--- a/packages/frontile/src/components/collections/dropdown.md
+++ b/packages/frontile/src/components/collections/dropdown.md
@@ -639,6 +639,10 @@ Menu items deliberately carry no `aria-selected`: it is invalid on a plain `menu
 conveys state through `aria-checked` and only as `menuitemcheckbox` or `menuitemradio`. A
 menu is a list of commands rather than a set of choices.
 
+Menu items share the Listbox's roving `tabindex`: exactly one item carries `tabindex="0"` —
+the active one, falling back to the first selected item and then to the first item that is not
+disabled — and every other item is `-1`.
+
 Keys are handled in two places, which is worth knowing when debugging one that doesn't fire:
 
 | Focus is on the trigger | Behavior                                          |
--- a/packages/frontile/src/components/collections/listbox.md
+++ b/packages/frontile/src/components/collections/listbox.md
@@ -546,6 +546,13 @@ export default class EmptySelection extends Component {
 | Items    | `role="option"` (or `menuitem`), `aria-labelledby` pointing at the item's label                |
 |          | `aria-selected` reflecting selection — options only, since it is invalid on a plain `menuitem` |
 |          | `aria-disabled="true"` for keys in `@disabledKeys`                                             |
+|          | a roving `tabindex` — exactly one option carries `0`, every other one `-1`                      |
+
+The options form a composite you step into once and then navigate with the arrow keys, so only
+one of them is ever in the tab order. That one is the active option; with nothing active it is
+the first selected option, and failing that the first option that is not disabled. A
+multiple-selection list with eight selections is therefore still a single stop for `Tab`, not
+eight.
 
 Keyboard, handled on the list itself:
 
--- a/packages/frontile/src/components/collections/listbox/item.gts
+++ b/packages/frontile/src/components/collections/listbox/item.gts
@@ -101,11 +101,14 @@ class ListboxItem extends Component<ListboxItemSignature> {
     }
   }
 
+  /**
+   * A roving tabindex: the manager nominates one option as the composite's tab
+   * stop and every other option stays out of the tab order, as the ARIA
+   * listbox pattern requires. Giving a `0` to each active *or selected* option
+   * turned a multi-select with eight selections into eight tab stops.
+   */
   get tabindex() {
-    if (this.listItem?.isActive || this.listItem?.isSelected) {
-      return 0;
-    }
-    return -1;
+    return this.manager.isTabStop(this.key) ? 0 : -1;
   }
 
   get classNames() {
--- a/packages/frontile/src/components/collections/listbox/listbox.gts
+++ b/packages/frontile/src/components/collections/listbox/listbox.gts
@@ -40,7 +40,7 @@ interface ListboxSignature<T> {
     elementToAddKeyboardEvents?: HTMLElement;
 
     /**
-     * @edefaultValue 'frist'
+     * @defaultValue 'first'
      */
     autoActivateMode?: 'none' | 'first' | 'selected';
 
@@ -106,7 +106,7 @@ class Listbox<T = unknown> extends Component<ListboxSignature<T>> {
       }
     } else {
       if (
-        ['Enter', ' ', 'Space'].includes(event.key) &&
+        ['Enter', ' '].includes(event.key) &&
         this.listManager.searchKeys == ''
       ) {
         this.listManager.selectActiveItem();
@@ -213,11 +213,17 @@ class Listbox<T = unknown> extends Component<ListboxSignature<T>> {
       tabindex="0"
       role={{this.role}}
       aria-multiselectable={{this.ariaMultiselectable}}
+      {{! The callbacks travel through setup, not only through the manager's
+      field initializer above: that initializer runs once, so an onAction
+      rebuilt on a later render would otherwise never reach the manager. }}
       {{this.listManager.setup
         selectedKeys=@selectedKeys
         disabledKeys=@disabledKeys
         selectionMode=@selectionMode
         allowEmpty=@allowEmpty
+        onAction=@onAction
+        onSelectionChange=@onSelectionChange
+        onActiveItemChange=@onActiveItemChange
         autoActivateMode=(if
           (isUndefined @autoActivateMode) "first" @autoActivateMode
         )
--- a/packages/frontile/src/utils/listManager.ts
+++ b/packages/frontile/src/utils/listManager.ts
@@ -1,8 +1,8 @@
 /* eslint-disable ember/no-runloop */
 import { tracked } from '@glimmer/tracking';
-import { debounce } from '@ember/runloop';
+import { cancel, debounce, later } from '@ember/runloop';
 import { modifier } from 'ember-modifier';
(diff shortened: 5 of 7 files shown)


josemarluedke:

The verdict I most wanted checked came back clean: **the load-bearing guards were respected, not bypassed.** `_tabStopKey` has its own mirror-guarded setter matching `ListItem`'s idiom; `#refreshTabStop` reads only `isActiveUntracked`/`isSelectedUntracked`/`isDisabledUntracked`, so it consumes no sibling tags on a path that can run mid-render; `unregister` deliberately doesn't clear the tab stop, with the hazard named in a comment; and `#orderedItems` is genuinely not memoized. Spec axis also confirmed the three cosmetics landed and `keypress` was **not** migrated.

One real regression, though — and it's in the thing this PR exists to improve.

### 🔴 Must fix — a perf regression inside the perf fix

The deleted `#firstVisibleItem` carried an explicit rationale: *"found in a single pass so callers that need only one item do not pay to order the whole list."* Its replacement iterates `#orderedItems`, and `#refreshTabStop()` is now called from `activateItem` (`listManager.ts:487-490`).

`setNextOptionActive`/`setPreviousOptionActive` already take `#orderedItems`, so **every arrow keypress now sorts the entire list twice** via `compareDocumentPosition`. On the 500-option list this PR is meant to make bearable, that's the same quadratic pressure moved from render into navigation.

**Fix:** derive the tab stop from the single item being activated rather than re-ordering the whole list — `activateItem` already holds it. Keep the ordered scan only for the fallback path (no active item), where it's genuinely needed.

### 🟡 Should fix — unrequested change, worth a line in the PR body at minimum

`listManager.ts:494-500`: `activateItem`'s `requestAnimationFrame` callback gained an `item.el.isConnected` guard. It's harmless and arguably a latent-bug fix, but it isn't in any finding, in the brief, or in the accepted-deviations list. Either call it out explicitly or split it.

- **Data clumps:** `#pendingAdd`/`#pendingRemove`/`#pendingAction` are three readings of one state — `#pendingAction` is truthy exactly when `#pendingAdd || #pendingRemove`, so the `if (action && typeof onListItemsChange === 'function')` check in `#flushListChange` reads it a third time. One `#pending?: { added, removed, last }` would say it once.
- `'onAction' in args` makes `updateArgs`' behaviour depend on key *presence*, invisible at the call site. The comment carries it, but it's a subtle contract for an exported util with three consumers.
- `#refreshTabStop` `continue`s on disabled items *before* the active check, so an active-but-disabled option would cede its tab stop. Unreachable through current activation paths.
- **Divergent Change was weighed and dismissed:** all six pieces are the same module's batch/tab-stop/callback plumbing, and the result reads as one coherent change.
- Coalescing tests assert `reports.length === 1` where `main` produces 8/2/5; the roving-tabindex tests use `document.querySelectorAll` rather than `assert.dom`, correctly avoiding the container-scope trap.

josemarluedke:
Both findings addressed.

**Finding 1.** You were right that the arrow path ordered the list twice. The observable is ordered scans, not wall clock: one scan reads `el.isConnected` exactly once per registered item, so a keypress's scan count is an exact, engine-independent integer. Over 50 items, one `setNextOptionActive()` read it **100 times (2 scans, 98 `compareDocumentPosition` calls)**. I saw that test fail on the branch as it stood (`actual: 2, expected: 1`) before touching the fix. After: **50 reads, 1 scan**.

The fix passes the just-activated item into `#refreshTabStop`, which names it as the tab stop directly. The ordered scan stays for the fallback path with no active item, and a disabled item still falls through to it; a second test pins that fallback at one scan.

Mirror guards untouched: `_tabStopKey`'s setter, `#refreshTabStop` reading only untracked mirrors, and `unregister` not clearing the stop all stand.

**Finding 2.** Disclosed both ways — a comment saying it is unrequested and why, in its own commit (`3ee592f`).

Full suite: **911 tests, 908 passed, 3 skipped, 0 failed.** `lint:types` clean.

PR #525, opened 2026-08-28
fix(forms): fire Select/Autocomplete @onBlur on real focus loss

josemarluedke:
## Problems

**1. `@onBlur` was synthesized from selection instead of from focus.** `Select.onAction` — which `ListManager.selectItem` calls on **every** item click — ran `later(() => this.args.onBlur?.(), 150)`. Nothing had blurred. `Autocomplete` had the same pattern.

Why it matters: `Field` wires `onBlur` to blur-validation. With `@selectionMode="multiple"` the dropdown stays open, so clicking the first of three options fired validation 150 ms later and painted "at least 3 required" under a field the user was still filling in. The `later` was also never cancelled, so it fired after teardown.

**2. The clear button bypassed `@allowEmpty`** — `clearSelectedKeys` called `onSelectionChange([])` with no check, while chips and listbox options both honour `canDeselectKey`.

## Fix: real focus tracking

A new `-private/control-blur.ts` (`ControlBlurTracker`) decides when a trigger-plus-dropdown control has genuinely lost focus, and reports it once.

- Listens to **`focusout`**, not `blur` — `relatedTarget` is only guaranteed on `focusout`.
- The control is two regions: the container element, and the popover content located via the trigger's `aria-controls` (the popover's trigger modifier sets it to the content id). That handles the portaling with no extra plumbing.
- **`relatedTarget` present → decide synchronously, no timer.** Covers Tab, clicks on other focusable things, and every move *into* the dropdown (an option click focuses the `li`, which has a real tabindex) — the case that used to fire spuriously.
- **`relatedTarget` null → one 150 ms `later`, then judge `document.activeElement`.** Focus nowhere *while the dropdown is open* is a click on the dropdown's own padding, not a blur; the same state with the dropdown closed is the user having clicked away. The 150 ms is the same "wait a beat" the old code used, here so an outside click has finished dispatching (and closed the popover) before state is read.
- **Teardown:** the single timer is cancelled in `willDestroy` on both components, and both the timer callback and the report path re-check `isDestroyed || isDestroying`.

`SelectTrigger`'s internal `@onBlur` became `@onFocusOut: (event: FocusEvent) => void`, wired with `on "focusout"`. `SelectTrigger` is not re-exported from `index.ts` or `forms.ts`, so this is internal.

## `@isClearable` vs `@allowEmpty`: kept as an override

`@isClearable` remains an explicit override of `@allowEmpty`, and code, docs and tests now all say so. Reasons: the existing `types.ts` JSDoc already documented it that way; `@allowEmpty` defaults to `false`, so honouring it would remove the clear button from every default configuration; and `@allowEmpty` governs deselecting *an option*, whereas the clear button is a separate opt-in affordance for emptying the field. A comment records the decision so it doesn't get "fixed" back.

The one genuinely dead button **was** removed: the clear button on a disabled control, which had no disabled state of its own and was focusable.

## Tests

12 added; 10 confirmed failing first.

Select: no `@onBlur` while selecting in multiple mode with the dropdown open; none when selecting in single mode; exactly once when focus leaves; the filterable variant (input trigger) — none while typing or selecting, once on leaving; teardown during `onSelectionChange` calls nothing; no clear button on a disabled Select. Autocomplete: the same set.

The two "exactly once" tests were **mutation-tested** — neutering `handleFocusOut` makes both fail, so they are not passing vacuously. No keyboard-activation assertions were added, so the `autoActivateMode` vacuous-pass trap doesn't apply here.

Two tests pass in both directions and are kept as guards: `@isClearable` clearing with `@allowEmpty={{false}}` (codifies the documented override) and Autocomplete not blurring while the dropdown is open.

Full suite: **909 tests, 906 pass, 3 skip, 0 fail** (897 baseline + 12 new). Filters: select 159 (1 pre-existing skip), autocomplete 33/33, form 492 (3 skips). `lint:types` clean. No pre-existing test was modified or deleted.

- **The 150 ms constant and the "focus nowhere + dropdown open ⇒ not a blur" rule.** A backdrop click is the case worth sanity-checking by hand.
- `ControlBlurTracker` lives in `-private/` deliberately. It first landed in `utils/`, which is listed in the package's `publicEntrypoints` — that would have published `frontile/utils/control-blur` as API we owe compatibility on. The trade-off is that the tracker has no direct unit-test path and is covered only through the Select/Autocomplete integration tests, which is where the coverage already was.
- **Known limitation:** if the overlay does not restore focus to the trigger after a single-mode selection, focus ends on `body` with no further `focusout` on the trigger, so no blur is reported until the user interacts again. This is reachable via the popover's permanent `preventFocusRestore` after a Tab-close — a separate pre-existing bug, independently found during the popover work and recorded there as a follow-up. In that path a blur had already been reported when the user tabbed out.
- Hiding the clear button when disabled is a small behavior change beyond the two stated bugs.

### DIFF
--- a/packages/frontile/src/-private/control-blur.ts
+++ b/packages/frontile/src/-private/control-blur.ts
@@ -1,0 +1,158 @@
+import { later, cancel } from '@ember/runloop';
+
+/**
+ * How long to wait before deciding that a focus loss with no identifiable
+ * destination really left the control.
+ *
+ * Only the targetless case waits at all (see {@link ControlBlurTracker}): a
+ * `focusout` that names where focus went is answered synchronously. Targetless
+ * blurs are the ones a click on something unfocusable produces -- the page
+ * background, a backdrop, the dropdown's own padding -- and the browser
+ * dispatches the rest of that click (mouseup, click) in later tasks, which is
+ * where an outside click gets around to closing the popover. Deciding before
+ * that would read a control that still looks open and stay silent. A beat is
+ * long enough for the click to finish and short enough that blur validation
+ * still feels immediate.
+ */
+const FOCUS_SETTLE_MS = 150;
+
+interface ControlBlurTrackerOptions {
+  /**
+   * The element the popover's `trigger` modifier is installed on. Its
+   * `aria-controls` is how the portaled content is found, so this must be the
+   * element that opens the dropdown, not merely something inside the control.
+   */
+  trigger: () => HTMLElement | null | undefined;
+
+  /** The control's outermost element: trigger, chips, clear button and all. */
+  container: () => HTMLElement | null | undefined;
+
+  /** Whether the control's dropdown is currently open. */
+  isOpen: () => boolean;
+
+  /** Whether the owning component has been (or is being) destroyed. */
+  isDestroyed: () => boolean;
+
+  /** Called once each time focus genuinely leaves the control. */
+  onBlur: () => void;
+}
+
+/**
+ * Decides when a composite control -- a trigger plus a dropdown -- has really
+ * lost focus, and reports it once.
+ *
+ * A trigger's own blur is not the answer. Clicking an option moves focus onto
+ * that option, which is *inside* the control as far as the user is concerned
+ * but is not a DOM descendant of the trigger: the popover content is portaled
+ * to the end of the document. And in multiple selection mode the dropdown
+ * stays open across several such clicks, so a trigger-blur-means-blur rule
+ * fires mid-interaction -- which is what made blur validation paint "pick at
+ * least three" under a field the user was still filling in.
+ *
+ * So the control is treated as two regions, the container and the popover
+ * content, and focus leaving one for the other is not a blur:
+ *
+ * - **`focusout` naming its destination** (`relatedTarget`) is answered
+ *   immediately, with no timer at all. This covers every deliberate exit --
+ *   Tab, or a click on another focusable thing -- and every move *into* the
+ *   dropdown, including the option click that starts a selection.
+ * - **A targetless `focusout`** is deferred by {@link FOCUS_SETTLE_MS} and then
+ *   judged on `document.activeElement`. Focus nowhere (`body`) while the
+ *   dropdown is open is still part of the interaction -- a click on the
+ *   dropdown's padding -- so it is not a blur; once the dropdown has closed,
+ *   the same state means the user clicked away.
+ *
+ * The portaled content is located through the trigger's `aria-controls`, which
+ * the popover's trigger modifier sets to the content's id. That needs no extra
+ * plumbing through the popover and stays correct wherever the content is
+ * rendered.
+ *
+ * Callers must call {@link cancel} from `willDestroy`: a pending timer would
+ * otherwise resolve against a torn-down component and both write tracked state
+ * and call a callback the consumer has stopped expecting.
+ */
+class ControlBlurTracker {
+  #timer?: ReturnType<typeof later>;
+
+  private readonly options: ControlBlurTrackerOptions;
+
+  constructor(options: ControlBlurTrackerOptions) {
+    this.options = options;
+  }
+
+  /**
+   * Wire this to `focusout` on the trigger -- not `blur`, whose
+   * `relatedTarget` browsers are not obliged to populate.
+   */
+  handleFocusOut = (event: FocusEvent): void => {
+    this.cancel();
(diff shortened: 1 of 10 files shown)


josemarluedke:

Placement prohibition verified: the helper is at `packages/frontile/src/-private/control-blur.ts`, `rollup.config.mjs` is untouched, and nothing new is published. Teardown is handled as asked — the single `later` is cancelled from `willDestroy` on both components, both the settle callback and `#report` re-check `isDestroyed || isDestroying`, and no `later` remains in either component file. The teardown-during-selection tests are real repros.

One behavioural gap, which my own browser verification observed and I mis-read as correct.

### 🔴 Must fix (or consciously accept + document) — multiple mode never blurs on click-away

The spec required: *"several option clicks in multiple mode with the dropdown open → no fire; tab/click away → fire once."* The no-fire half is covered. **The fire-once half is only tested in single mode and in the `@isFilterable` variant** — both cases where the popover has already closed and restored focus to the trigger.

In multiple mode the dropdown stays open and focus sits on the option `<li>` inside the *portaled* content, while the `focusout` listener is installed only on the trigger/input (`trigger.gts:125, :143`). I confirmed the consequence by hand in a browser:

| action | `@onBlur` calls |
|---|---|
| click option "Ada" (dropdown stays open) | 0 ✅ |
| click option "Grace" | 0 ✅ |
| **click an outside button** → popover closes, Overlay restores focus to trigger | **0** ⚠️ |
| click that outside button *again* | 1 |

So after a multi-select interaction, clicking away does **not** fire `@onBlur` — the user must interact a second time. Since `Field` wires `@onBlur` to blur-validation, that means blur-validation no longer runs when a user finishes with a multi-select and clicks away. The old bug fired too eagerly; this is the mirror image.

Strictly speaking focus *is* on the trigger, so the tracker is behaving per its own rules — the trigger is the Overlay's focus-restore target. That makes this arguably an Overlay question rather than a tracker bug, which is why I'm flagging it for a decision rather than dictating a fix.

**Please do one of:**
1. Treat it as a bug — don't restore focus to the trigger when the close was caused by a click on another focusable element; or
2. Consciously accept it, add a test pinning the *actual* behaviour so it can't drift silently, and document it in `select.md` next to the `@onBlur` semantics.

Either way the multiple-mode exit path needs a test, and there is currently **no keyboard exit test at all** — every added exit test uses `click`, so the `relatedTarget`-present synchronous branch is never exercised via `Tab`.

- **Duplicated wiring:** the five-closure `new ControlBlurTracker({...})` block plus the `willDestroy` cancel and JSDoc are copied verbatim into `select.gts` and `autocomplete.gts`. A `trackControlBlur(this)` factory would collapse both and make the "callers must cancel in `willDestroy`" contract impossible to forget — which is the failure mode the class's own JSDoc warns about.
- The `@allowEmpty`-override rationale is now written out four times (`types.ts`, `select.gts`, `select.md`, `autocomplete.gts`), and one copy admits it: *"the two must keep saying the same thing."* One canonical place (the `@isClearable` JSDoc, which feeds the generated API table) plus pointers.
- `control-blur.ts` checks `isDestroyed()` in both `#settle` and `#report`; `#settle`'s only exits lead to `#report`, so one is dead.
- The standards reviewer flagged the `isClearable` change as scope creep. It isn't — it was Finding 2 in the brief. Noting so the record is straight.

josemarluedke:
**Reproduced.** Multiple mode, two option clicks, then an outside click expecting one `@onBlur`: `actual: 0, expected: 1`. Your browser observation was right. The outside click is spent closing the dropdown, and the Overlay restores focus to the trigger as it closes, so focus is still inside the control when the `focusout` settles — `ControlBlurTracker` is behaving per its own rules.

**Pinned rather than fixed.** The real fix is in the Overlay's focus-restore, and `overlay.gts` belongs to #519, so changing it here would collide. Instead there's now a test named `in multiple mode the trigger keeps focus after an outside click, so no blur is reported until focus truly leaves`, asserting 0 blurs after the closing click and exactly 1 after the next one. Both halves are pinned, so neither can drift silently. `select.md` states the one-interaction delay next to the `@onBlur` semantics.

**Keyboard exit covered.** Every existing exit test used `click` (targetless, deferred branch). Added Tab-away tests for Select and Autocomplete that take the synchronous `relatedTarget` branch — both assert exactly one `@onBlur`, both green.

Full suite: **912 tests, 909 pass, 3 skip, 0 fail.** `lint:types` clean.

PR #526, opened 2026-09-03
fix(theme): correct body text-scale font sizes

josemarluedke:
## Summary
- The `--text-body-*` tokens in the typography scale were mapped to the wrong `--font-size-N` steps, causing every size from `xs` up through `xl` to render larger than the design spec.
- Remapped each body size token to the correct modular-scale index, matching the designer-provided spec (px) report.
- Added the missing `4xs`, `5xs`, `2xl`, and `3xl` body sizes to fill out the scale, and registered them in `COMMON_UNITS` (tailwind-merge conflict resolution).
- Updated the co-located typography docs demo and CSS variables reference doc to reflect the new/corrected sizes.

## Test plan
- [x] `pnpm --filter @frontile/theme build`
- [x] `pnpm --filter frontile build`
- [x] `pnpm --filter @frontile/theme lint:types` / `pnpm --filter frontile lint:types`
- [x] `pnpm lint:js` (no new errors; pre-existing warnings only)
- [x] Verified in-browser on the docs typography page that computed `font-size` for every `text-body-*` class matches spec exactly (5xs 8.37px → 3xl 28.68px)

### DIFF
--- a/docs/migrations/v0.18/index.md
+++ b/docs/migrations/v0.18/index.md
@@ -24,6 +24,8 @@ them is optional for the whole 0.18 line.
 | `@frontile/*` package imports | Nothing — they still work in 0.18.x | Deprecation warning only |
 | Derived border-radius scale | Slightly rounder corners on menus and small marks | Visual only — nothing to fix |
 | Multi-select renders chips | Multi-selects look different — selections become removable chips | Visual only — nothing to fix |
+| `text-body-pico`/`-nano`/`-micro` | Those elements render unstyled | **Silently** |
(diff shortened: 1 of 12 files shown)


voguetheory (approved):
I can't seem to find the pixel values to verify sizes but I trust that it matches the scale. (Confirming the naming convention for nano/pico were replaced with the applicable ([number]xs) values.)
Attaching the styles export that Claude referenced earlier, (which I have also double-checked) so as long as that matches I think we're good on body.

[typography-styles.json](https://github.com/user-attachments/files/31800814/typography-styles.json)

PR #540, opened 2026-09-05
feat(site): move every code block to Docfy's Shiki pipeline

josemarluedke:
> **Draft: blocked on a Docfy release.** This branch needs `@docfy/plugin-shiki`, which is not published yet, and Docfy changes that are still in review (josemarluedke/docfy#222). See *Before this can merge* at the bottom.

Moves every code block on the site to Docfy's new Shiki pipeline.

## The site had three separate highlighters

All three are gone, and **nothing tokenises in the browser any more**.

| | Was | Now |
| --- | --- | --- |
| Docs fences | `rehype-highlight` + a hand-written `glimmerTypescript` wrapper over `highlightjs-glimmer` | `@docfy/plugin-shiki` |
| Homepage code panel | `lowlight` + 7 highlight.js grammars, **in the browser** | pre-highlighted at build time |
| API signature tables | `lowlight` in a Node script | the same Shiki preset as the docs |

Auditing for this turned up four touch points beyond the obvious three: `app/styles/highlight.css` (including an `.hljs-light-theme` scope), two `hljs-light-theme` usages plus an `hljs-name` span in `signature.gts`, `docfy-demo.css` painting snippet `pre`s with prose's always-dark background, and the signature-markdown test fixtures.

## The homepage got smaller, not bigger

The panel highlighted six fixed strings at runtime, which put lowlight, seven grammars and `highlightjs-glimmer` into the bundle every visitor downloads — 64.7 KB minified, 20.5 KB gzipped. They are baked at build time now by a generator beside the existing `generate-signature-data` one, and the generated module costs 2.4 KB gzipped. **The homepage is roughly 18 KB lighter.**

The theme lab's ramp derivation moved into that generator rather than being duplicated, so the live panels and the config snippet beside them can no longer disagree.

## What this buys the docs

`gts` is the majority language across Frontile's documentation, and Shiki ships real `glimmer-js` / `glimmer-ts` grammars — so `<template>` tags and `{{mustaches}}` are tokenised as themselves instead of falling back to plain JavaScript. That is the main reason for the move, and it retires the hand-written `glimmerTypescript` wrapper.

Every fence is now wrapped in `DocfyCodeBlock`, which brings a copy button, `title=`, `collapsible`, `{a-b}` line ranges and `:::code-tabs`. The install instructions, the file-path examples and the two longest config samples use them.

Themes are emitted as `--shiki-light` / `--shiki-dark` custom properties, so light/dark switching needs no re-highlighting and no JavaScript. `app/styles/highlight.css` (fixed GitHub Dark colours) is replaced by `docfy-code-block.css`, which styles Docfy's BEM classes with Frontile's own tokens.

## Verification

- `pnpm --filter site test` — 23/23. `pnpm --filter site test:node` — 30/30.
- Client and SSR builds pass.
- Rendered output checked directly: the prerendered homepage carries `<code class="shiki" data-language="…">` with real `--shiki-light`/`--shiki-dark` token spans and zero `hljs`; compiled docs chunks carry `data-language="glimmer-ts"` and the `:::code-tabs` labels.
- Chrome measured in both themes: card/border/token colours, and `--docfy-code-block-fade` equal to the card background so the collapse fade blends. This caught a real bug during development — an earlier revision used `bg-surface-canvas` and the border vanished in light mode.

## Before this can merge

1. **Docfy must ship** josemarluedke/docfy#222, which adds `@docfy/plugin-shiki` and the code-block components.
2. **`pnpm-lock.yaml` needs regenerating** once those are on npm. It is intentionally unchanged here: it does not yet carry the `@docfy/plugin-shiki` entry, because the package does not exist to resolve.
3. **CI will not install until then**, for the same reason.

One local-only caveat for anyone picking this up: verifying it requires `link:` overrides to the unpublished Docfy packages, and those break the SSR **prerender** step (`Cannot read properties of undefined (reading 'children')` on docs routes). That is an artifact of linking, not a defect in this branch — I bisected it: the same site with *published* Docfy prerenders 68/68, and the published versus source-built package files are byte-identical. The dev server renders correctly throughout. It resolves itself once the packages are real.


josemarluedke:
Unblocked: Docfy **0.14.0** is published, so this branch now builds against the released packages instead of local `link:` overrides.

- `site/package.json` moved to `^0.14.0` for all five `@docfy/*` packages, and `pnpm-lock.yaml` is regenerated against the registry — it now carries the `@docfy/plugin-shiki` entry it previously could not resolve.
- Full site build passes end to end: **75 rendered, 0 failed** (client, SSR, prerender).
- `pnpm --filter site test` 23/23, `test:node` 30/30.

This also settles the prerender failures noted during review. They were an artifact of `link:`-ing unpublished packages, not a defect in this branch — the same site prerendered 68/68 with published Docfy while the linked, byte-identical source build failed 67. Against the real packages it passes.

Marking ready for review.

PR #552, opened 2026-09-07
fix(overlays): forward preventAutoFocus from Drawer and Modal to Overlay

### LINKED ISSUE #551: Drawer and Modal don't forward Overlay's preventAutoFocus
## What's wrong

`Overlay` supports `@preventAutoFocus` to opt out of its auto-focus-on-open behavior when
`@disableFocusTrap={{true}}` is set. But `Drawer` and `Modal` — the two components most people
actually use — don't forward it: it's missing from both `DrawerArgs`/`ModalArgs`' `Pick<...>`
of `Overlay`'s args, and from their `<Overlay @... >` invocation in the template. So there's no
way to reach this option from the public API of either component.

## Why it matters

A common pattern for a non-modal, dismissable overlay is `@disableFocusTrap={{true}}`: the page
behind stays interactive, focus isn't forced into the dialog, and the user can keep doing
whatever they were doing while the dialog is present (e.g. an inline "assistant" panel, a
persistent info drawer). But `Overlay`'s own focus trap opt-out still auto-focuses the dialog's
root element ~200ms after it opens (see `setupContent` in `overlay.gts`) — the trap is off, but
`disableFocusTrap` isn't the same thing as "don't touch focus at all". `preventAutoFocus` is
exactly the escape hatch for that, but it's stuck on `Overlay` with no way to set it through
`Drawer`/`Modal`.

Concretely, given something like:

```gts
<Drawer @isOpen={{this.isOpen}} @disableFocusTrap={{true}} as |d|>
  <d.Body>...</d.Body>
</Drawer>
```

there's no `@preventAutoFocus` to add — the auto-focus always happens, which can be surprising
(scrolls the page, steals focus from whatever the user was interacting with) for a dialog that
was deliberately built to not take over the page.

## What we're trying to achieve

Forward `preventAutoFocus` through `Drawer` and `Modal` the same way `disableFocusTrap` and the
other `Overlay` args already are, so:

```gts
<Drawer
  @isOpen={{this.isOpen}}
  @disableFocusTrap={{true}}
  @preventAutoFocus={{true}}
  as |d|
>
  <d.Body>...</d.Body>
</Drawer>
```

opts fully out of both the focus trap and the auto-focus, leaving focus wherever it already was.

MichalBryxi:
## Summary

- `Overlay` already supports `@preventAutoFocus` to opt out of its auto-focus-on-open behavior when `@disableFocusTrap={{true}}`, but `Drawer` and `Modal` never forwarded it — it was missing from both `DrawerArgs`/`ModalArgs`' `Pick<...>` of `Overlay`'s args, and from their `<Overlay>` invocation.
- Adds `preventAutoFocus` to both components' picked args and forwards it to their internal `<Overlay>`, matching how every other `Overlay` arg is already forwarded.
- Updates the co-located `drawer.md`/`modal.md`/`overlay.md` docs to mention the auto-focus-without-a-trap behavior and the new escape hatch.

Fixes #551.

## Test plan

- [x] Added a failing-first test to both `drawer-test.gts` and `modal-test.gts` asserting the dialog auto-focuses by default when `@disableFocusTrap={{true}}`, and that `@preventAutoFocus={{true}}` prevents it — confirmed each fails without the fix and passes with it.
- [x] `pnpm --filter @frontile/theme build && pnpm --filter frontile build`
- [x] `cd test-app && pnpm ember test --filter="Drawer"` (28/28 pass), `--filter="Modal"` (32/32), `--filter="Overlay"` (109/109)
- [x] `pnpm exec eslint` / `pnpm exec ember-template-lint` on the changed files — clean
- [x] `pnpm --filter frontile lint:types` — clean
- [x] `cd test-app && pnpm lint:types` — no new errors (diffed against `main`; the pre-existing failures in unrelated files are unchanged)

### DIFF
--- a/packages/frontile/src/components/overlays/drawer.gts
+++ b/packages/frontile/src/components/overlays/drawer.gts
@@ -36,6 +36,7 @@ export interface DrawerArgs extends Pick<
   | 'closeOnOutsideClick'
   | 'closeOnEscapeKey'
   | 'backdropTransition'
+  | 'preventAutoFocus'
 > {
   /**
    * The transition to be used in the Drawer.
@@ -234,6 +235,7 @@ export default class Drawer extends Component<DrawerSignature> {
       @backdropTransition={{@backdropTransition}}
       @transition={{this.transition}}
       @closeOnOverlayElementClick={{true}}
+      @preventAutoFocus={{@preventAutoFocus}}
     >
       <div
         class={{this.classes.base class=@classes.base}}
--- a/packages/frontile/src/components/overlays/drawer.md
+++ b/packages/frontile/src/components/overlays/drawer.md
@@ -736,7 +736,9 @@ In development, a drawer that ends up with no accessible name at all — no `Hea
 `frontile.drawer.missing-accessible-name`. It is compiled out of production builds.
 
 `aria-modal="true"` is dropped when `@disableFocusTrap={{true}}`: with the trap off the page
-behind really is reachable, and claiming otherwise would mislead screen reader users.
+behind really is reachable, and claiming otherwise would mislead screen reader users. Note
+that the drawer still auto-focuses itself in this case, unless `@preventAutoFocus={{true}}`
+is also passed — see [Overlay](./overlay.md#accessibility).
 
 Behavior inherited from [Overlay](./overlay.md):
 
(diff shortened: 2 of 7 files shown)


josemarluedke (approved):

josemarluedke:
Released under v0.18.0-alpha.18.

PR #586, opened 2026-09-15
fix(collections, overlays): menu selection semantics, keyboard open, focus trap defaults

josemarluedke:
Four fixes found while building a faceted filter menu (Linear-style property/operator/value pills) on `Dropdown`. Each is small and independent; they are together because they were all in the way of the same UI.

## 1. Selectable menu items announced no state

`Listbox` rendered every menu row as a plain `menuitem` with no `aria-checked`, so a menu with a `@selectionMode` sounded identical to a list of commands. `item.gts` already documented the correct behaviour in a comment — menus convey state through `aria-checked`, and only as `menuitemcheckbox`/`menuitemradio` — but `get role()` always returned `menuitem`.

Rows now take the matching checkable role and report `aria-checked`. A row that opens a submenu stays a plain `menuitem`: activating it opens a menu rather than toggling a value. Both `role` and `ariaChecked` hang off one `isCheckableMenuRow` getter, so they cannot disagree about whether a row carries state.

`ListManager` gains a public `selectionMode`, since the row is handed the manager rather than the mode. It is **tracked**, written through the mirror guard `ListItem` documents: read plain, a row's `role` kept its old value when a consumer flipped `@selectionMode` at runtime while the item's own tracked `isSelected` moved on — which lands `aria-checked` on a plain `menuitem`, the exact invalid ARIA the checkable roles exist to avoid. `updateArgs` also assigned the mode only when truthy, so `undefined` could never reset it; it now defaults to `'none'` like its neighbours.

## 2. Opening a menu from the keyboard landed on nothing

The WAI-ARIA menu button pattern opens onto an item when the keyboard opened the menu. The root menu always used `autoActivateMode: 'none'`, so the first `ArrowDown` after opening was spent getting to the first row.

The trigger now reports how it was reached, using the `OpenSource` type `menu-context.ts` already exports and `Sub` already uses — rather than a second spelling of the same idea — and the root menu activates its first row only for a keyboard open. A pointer open still highlights nothing, and an explicit `@autoActivateMode` still wins. The trigger binds `pointerup` rather than `pointerdown`: it satisfies `no-pointer-down-event-binding` and records the source just as reliably, since `pointerup` precedes the `click` Popover opens on.

`ArrowUp` opening onto the *last* item is not included: `AutoActivateMode` has no `'last'`, and adding one touches several places in `ListManager`. Happy to follow up separately.

## 3. A submenu could not declare its own selection

Selection settings are declared on the root and inherited by every depth, which is the right default and well documented in `menu-context.ts`. But a submenu's own settings were silently ignored while still type-checking — they are in `MenuArgs`, so nothing tells you they do nothing.

All seven inheritable settings — `@selectionMode`, `@selectedKeys`, `@disabledKeys`, `@allowEmpty`, `@onAction`, `@onSelectionChange` and `@closeOnItemSelect` — now override the inherited context for that level, so a navigation menu can hold a multi-select submenu without the root pretending to select. Appearance and timing (`@variant`, `@color`, `@shortcutVariant`, `@disableTransitions`, `@transitionDuration`) stay the root's for the whole chain, so one menu cannot end up looking like two. A submenu that declares nothing inherits exactly as before, and is handed the parent's own context object back untouched.

`get context()` is `@cached`, for the same reason `variant` already is: it is read a dozen times per render, and the overriding branch builds a new object each call — so without the cache every read handed `Sub` a fresh identity to build a fresh child context from.

## 4. `@focusTrapOptions` replaced the defaults instead of merging

`Overlay` did `this.args.focusTrapOptions || { clickOutsideDeactivates: true, allowOutsideClick: true }`. Naming any single option — a `setReturnFocus`, an `initialFocus` — silently dropped both defaults, after which the focus trap swallowed outside clicks and the overlay could only be dismissed with Escape. Nothing warns, and the cause sits nowhere near the symptom.

Consumer options are now merged over the defaults. Passing either as `false` still opts out deliberately.

## Testing

- **12 new tests** — 10 in `dropdown-test`, 1 in `overlay-test`, 1 in `listbox-test` — each written failing first.
- One existing assertion updated: `listbox-test`'s "menu items do not carry aria-selected" pinned the old gap (`role="menuitem"`, no state at all). It now asserts the end state its own comment describes — still no `aria-selected`, but `menuitemradio` + `aria-checked`.
- Full suite: **1946 passing, 0 failing** (3 skipped).
- `lint:hbs` is one error better than before: the `pointerdown` binding was the only lint error this branch introduced, and it is gone.
- Docs updated in `dropdown.md` (selection roles, keyboard open, the full override list), `listbox.md` (the accessibility table described the old menu-item behaviour) and `overlay.md` (focus trap merge, `setReturnFocus`).

All four were verified in a real consuming app as well as in tests.


packages/frontile/src/components/collections/dropdown/dropdown.gts
```diff
@@ -367,6 +422,46 @@ class Menu extends Component<MenuSignature> {
...
 
+  /**
+   * Layers this level's own arguments over the inherited context.
+   *
+   * Only arguments actually written on this level count: reading them
+   * unconditionally would overwrite inherited values with `undefined` and
+   * silently break inheritance, which is the default and by far the common
+   * case.
+   */
+  #withOwnOverrides(inherited: MenuContext): MenuContext {
```
josemarluedke: Would this be reactive if any arg change? Does it need to be reactive?

